PR #10, opened 2022-08-23
specify --cache-from when building

pkwarren:
Use the ghcr.io images with --cache-from to optimize build times.


pkwarren:
This appears to do what we want - took ~6m56s to build using `--cache-from` vs. ~31m without.

mfridman (approved):
Great job! This is a huge improvement.

PR #11, opened 2022-08-23
use specific base images

pkwarren:
To prevent unnecessary building when base images change (and improve
build reproducibility), target the most specific version of base images.


mfridman (approved):
I presume we'll lean on dependabot to bump these? Will it be too noisy?

PR #15, opened 2022-08-23
Fix fetch versions

mfridman:
This PR updates the directory logic to account for nested subdirectories. E.g.,

```
Normal:
library/{plugin_name}/{version}
Example: library/connect-go/v0.1.1

Alternative:
library/{plugin_base}/{version}/{plugin_language}
library/grpc/v1.48.0/ruby
```

In the alternative format we capture the language to group common plugins together. There is a shared base directory for `library/grpc` and `library/protoc`, and this is more for convenience.


cmd/fetcher/main.go
```diff
@@ -1,28 +1,23 @@
...
+	"github.com/bufbuild/buf/private/pkg/encoding"
 	"github.com/bufbuild/plugins/internal/fetchclient"
 	"github.com/bufbuild/plugins/internal/source"
 	"go.uber.org/multierr"
 	"golang.org/x/mod/semver"
 )
 
-var (
-	// TODO(mf): use the plugin from bufbuild/buf
-	re = regexp.MustCompile(`(?m)^plugin_version:\s(v.*)$`)
```
mfridman, line 23: Removed this in favor of the `bufpluginconfig` config.

pkwarren (approved):

PR #26, opened 2022-08-23
run buf alpha plugin push on CI workflow

pkwarren:
Update the CI workflow (when changes are pushed to the main branch) to
run the `buf alpha plugin push` command (in dependency order) to push
all plugins to the BSR.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -77,7 +77,7 @@ push: build
 			CACHE_ARGS=" --cache-from $(DOCKER_CACHE_ORG)/plugins-$${PLUGIN_OWNER}-$${PLUGIN_NAME}:$${PLUGIN_VERSION}"; \
 		fi; \
 		if [[ "$(DOCKER_ORG)" != "bufbuild" ]]; then \
-			docker pull $(DOCKER_ORG)/bufbuild/plugins-$${PLUGIN_OWNER}-$${PLUGIN_NAME}:$${PLUGIN_VERSION} || :; \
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -69,6 +69,16 @@ test:
...
 	@for plugin in $(PLUGIN_YAML_FILES); do \
 		plugin_dir=$$(dirname $${plugin}); \
+		PLUGIN_FULL_NAME=$(shell yq '.name' $${plugin_dir}/buf.plugin.yaml); \
+		PLUGIN_OWNER=`echo "$${PLUGIN_FULL_NAME}" | cut -d '/' -f 2`; \
+		PLUGIN_NAME=`echo "$${PLUGIN_FULL_NAME}" | cut -d '/' -f 3-`; \
+		PLUGIN_VERSION=$(shell yq '.plugin_version' $${plugin_dir}/buf.plugin.yaml); \
+		if [[ -n "$(DOCKER_CACHE_ORG)" ]]; then \
+			CACHE_ARGS=" --cache-from $(DOCKER_CACHE_ORG)/plugins-$${PLUGIN_OWNER}-$${PLUGIN_NAME}:$${PLUGIN_VERSION}"; \
+		fi; \
+		if [[ "$(DOCKER_ORG)" != "bufbuild" ]]; then \
```
mfridman, line 79: The only thing that worries me here is if we can't pull the image we'll end up bumping the revision for _all_ pugins. Do you think we should fail hard if for whatever reason docker pull fails? I.e., `not || :;`

mfridman (approved):

mfridman (approved):

PR #29, opened 2022-08-23
update docker pull command

pkwarren:
Don't append 'bufbuild' to docker pull command and fail push if we can't
pull the image from the registry. Consistently use $(DOCKER) and add a
$(BUF) command as well for easy testing.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,6 +9,7 @@ DOCKER_CACHE_ORG ?=
 
 GO_TEST_FLAGS ?= -race -count=1
 
+BUF ?= buf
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -9,6 +9,7 @@ DOCKER_CACHE_ORG ?=
 
 GO_TEST_FLAGS ?= -race -count=1
 
+BUF ?= buf
```
pkwarren, line 12: By adding these you can quickly test the makefile by running `make push BUF="echo buf" DOCKER="echo docker"` to see the commands that would be executed.

mfridman (approved):

PR #34, opened 2022-08-25
optimize image publishing with --image flag

pkwarren:
Update the 'buf alpha plugin push' command to use the new '--image' flag
to publish a prebuilt image instead of repeating the build again.


Makefile
```diff
@@ -63,7 +63,13 @@ test:
 		CACHE_ARGS=" --cache-from type=registry,ref=$(DOCKER_CACHE_ORG)/plugins-$${PLUGIN_OWNER}-$${PLUGIN_NAME}:$${PLUGIN_VERSION}"; \
 	fi; \
 	test -n "$${PLUGIN_NAME}" -a -n "$${PLUGIN_VERSION}" && \
-	$(DOCKER) $(DOCKER_BUILD_ARGS) $(DOCKER_BUILD_EXTRA_ARGS)$${CACHE_ARGS} --build-arg PLUGIN_VERSION=$${PLUGIN_VERSION} -t $(DOCKER_ORG)/plugins-$${PLUGIN_OWNER}-$${PLUGIN_NAME}:$${PLUGIN_VERSION} $(<D)
+	$(DOCKER) $(DOCKER_BUILD_ARGS) \
+		$(DOCKER_BUILD_EXTRA_ARGS)$${CACHE_ARGS} \
+		--build-arg PLUGIN_VERSION=$${PLUGIN_VERSION} \
+		--label build.buf.plugins.config.owner=$${PLUGIN_OWNER} \
+		--label build.buf.plugins.config.name=$${PLUGIN_NAME} \
```
pkwarren, line 70: Added labels to match `buf alpha plugin push`. Going to remove the remote label from the published images since we use `--override-remote` and other things which can change what is in the buf.plugin.yaml.

.github/workflows/deploy.yml
```diff
@@ -51,9 +51,7 @@ jobs:
           PLUGIN_OWNER=$(echo "${PLUGIN_FULL_NAME}" | cut -d '/' -f 2)
           PLUGIN_NAME=$(echo "${PLUGIN_FULL_NAME}" | cut -d '/' -f 3-)
           PLUGIN_VERSION=$(yq '.plugin_version' ${PLUGIN_YAML})
-          docker pull ghcr.io/bufbuild/plugins-${PLUGIN_OWNER}-${PLUGIN_NAME}:${PLUGIN_VERSION} || :
+          docker pull ghcr.io/bufbuild/plugins-${PLUGIN_OWNER}-${PLUGIN_NAME}:${PLUGIN_VERSION}
```
mfridman, line 54: If we can't pull an image (either it doesn't exist or some other error), won't this fail with exit code 1?

pkwarren: Yes - that's what we want. If we can't pull the image the `--image` flag won't work below.

mfridman: Okay perfect, just confirming this is the intended behavior.

mfridman (approved):

Makefile
pkwarren: https://github.com/bufbuild/buf/pull/1362 updates labels when building an image to match.

PR #37, opened 2022-08-26
Remove pushing all plugins

mfridman:
Continue to build and test _all_ plugins on merge to main, but do not do a subsequent push.


.github/workflows/ci.yml
```diff
@@ -44,12 +44,3 @@ jobs:
     - name: Test
       shell: bash
       run: make test DOCKER_ORG="ghcr.io/bufbuild"
-    - name: Push
```
mfridman, line 47: I can also comment this out and add a note if that's preferable.

pkwarren (approved):

PR #48, opened 2022-08-29
add an optional flag to generate to a subpackage

pkwarren:
When using remote generation, it is required to generate code to a
separate package. Add a new flag '--separate_package' to the
protoc-gen-go-grpc plugin to enable generating code to a subpackage
(similar to connect-go).


smallsamantha (approved):
very nice!

PR #50, opened 2022-08-29
Revert "Remove pushing all plugins (#37)"

pkwarren:
This reverts commit 19f40837e4d88e172b8025e2b0a0c698c33a29d3.


pkwarren:
We've fixed the issues we were seeing with unreproducible builds and would like to re-enable the publishing of plugins to the BSR.

PR #52, opened 2022-09-01
Add custom registry option

mfridman:
Custom options these 2 plugins support to output packages to separate directories, when used with the BSR's npm registry.


library/connect-web/v0.2.0/buf.plugin.yaml
```diff
@@ -6,6 +6,7 @@ description: Simple, reliable, interoperable. A better gRPC-web.
 deps:
   - plugin: buf.build/library/protobuf-es:v0.1.0
 registry:
+  rewrite_import_path_suffix: connectweb.js
```
pkwarren, line 9: Is this supported by earlier versions too or just the latest?

pkwarren (approved):

PR #53, opened 2022-09-01
add plugins for gRPC v1.48.1 release

pkwarren:


library/grpc/v1.48.1/base/Dockerfile
```diff
@@ -0,0 +1,9 @@
+# syntax=docker/dockerfile:1.4
+ARG DOCKER_ORG=bufbuild
+ARG BASE_BUILD_VERSION=latest
+FROM ${DOCKER_ORG}/plugins-grpc-base-build:${BASE_BUILD_VERSION}
+
+RUN git clone --depth 1 --branch v1.48.1 --recursive https://github.com/grpc/grpc
```
mfridman, line 6: Can you remind me, why did we hardcode the version here?

I presume it was to resolve a caching issue, but are we sure this is still needed ?

pkwarren: Yep - without this we'd get cache misses where the only difference was a created timestamp between two builds.

mfridman (approved):

PR #56, opened 2022-09-02
Add output_languages to all files

mfridman:
This PR updates `/library` plugins and categorized each plugin with `output_languages`.

### DIFF
--- a/contrib/chrusty/jsonschema/v1.3.9/buf.plugin.yaml
+++ b/contrib/chrusty/jsonschema/v1.3.9/buf.plugin.yaml
@@ -1,6 +1,6 @@
 version: v1
 name: buf.build/contrib/chrusty-jsonschema
 plugin_version: v1.3.9
-output_languages: [javascript]
+output_languages: []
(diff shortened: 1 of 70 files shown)


contrib/chrusty/jsonschema/v1.3.9/buf.plugin.yaml
```diff
@@ -1,5 +1,6 @@
 version: v1
 name: buf.build/contrib/chrusty-jsonschema
 plugin_version: v1.3.9
+output_languages: [javascript]
```
pkwarren, line 4: Not sure if we should categorize this as JavaScript. Maybe we need json api or open api or just to leave off language here (will fall into misc/other category?)

pkwarren (approved):

contrib/chrusty/jsonschema/v1.3.9/buf.plugin.yaml
mfridman: Ye, I like that. Let's leave it out for now. `output_languages: []`

PR #65, opened 2022-09-07
update README and add CONTRIBUTING.md

pkwarren:
Update the README to include basic information and instructions on requesting new plugins. Add a new CONTRIBUTING.md file with details on development (including steps remaining to be automated and more details on plugin authoring).


CONTRIBUTING.md
```diff
@@ -0,0 +1,113 @@
...
+We've only included a representative set of Docker images in the [.github/dependabot.yml](https://github.com/bufbuild/plugins/blob/main/.github/dependabot.yml) file for each type of plugin (Node, Go, etc.).
+If Dependabot detects an update to a base image, we should update all plugins to use the new base image either with a separate PR or a manual update to the Dependabot PR.
+
+## Local Testing
+
+When testing locally, you may wish to build for a different architecture or push plugins to a different instance of the BSR.
+For example:
+
+```
+$ make push DOCKER_ORG="bufbuild.internal" BUF_PLUGIN_PUSH_ARGS="--override-remote bufbuild.internal"
```
pkwarren, line 110: Still working to get this working locally. I'd like to use consistent ways to build in both CI and locally. Right now we're still using `buf alpha plugin push` support for building Docker images locally.

CONTRIBUTING.md
```diff
@@ -0,0 +1,113 @@
...
+### Creating a new plugin version
+
+If the `fetcher` command opens a PR for a new version of an existing plugin, there are currently some manual steps required before merging:
+
+1. Check out the PR locally (i.e. `gh pr checkout <num>`).
+2. Perform any manual updates to `package.json`, `go.mod`, or other files (if used by the plugin).
+   If NPM or Go dependencies change, ensure that `npm i` or `go mod tidy` are run to ensure that the `package-lock.json` or `go.sum` files are kept up to date.
+3. Double check version numbers in `plugin.yaml` or other files to ensure they are bumped.
+   Often times, there may be dependencies (plugin or runtime dependencies) which need to be updated.
+4. Run `make` and `make test` to build the plugin and run the unit tests.
```
mfridman, line 93: This is the part I've been struggling with lately. I'd love if we could avoid rebuilding the world when we want to work/commit a single plugin version.

Ideally we could target a specific versioned directory, which I think is possible with some of the env variables. If yes, should these instructions be updated?

Iirc the hesitation was that we open _multiple_ plugin versions updates per PR?

mfridman (approved):
lgtm. We can iterate on it as we go.

Nit, should the README have a link / section to CONTRIBUTING ?

CONTRIBUTING.md
mfridman: Tbh I wouldn't worry about this too much, and if it adds complexity I'd say the tradeoff isn't worth it

We rebuild clusters and prune images quite often, so building/pushing all these plugins (which will grow over time) to the local cluster isn't something folks will be doing because it decreases the development feedback loop.

I think it's enough to be able to clone this repository, and with `--override-remote` target the local cluster to upload specific plugin(s).

pkwarren:
> Nit, should the README have a link / section to CONTRIBUTING ?

I was hoping to keep the README simple to start but can add a note if it would be helpful.

CONTRIBUTING.md
pkwarren: If you've previously built everything it *should* only rebuild new/changed plugins on subsequent calls to make but I agree it would be helpful to target which plugins to build by name for example. I'll see what I can do here.

PR #66, opened 2022-09-07
always use multi-stage builds for node plugins

pkwarren:
Update node plugins to use a multi-stage build even though we're using package.json/package-lock.json to control dependency versions. There still might be state stored during the `npm ci` operation that we don't want to exist in the final runtime image.

This doesn't make a huge difference but it does reduce the image size by 2-3mb and will hopefully remove additional places where we'll have Docker image build cache misses.

### DIFF
--- a/library/connect-web/v0.0.10/Dockerfile
+++ b/library/connect-web/v0.0.10/Dockerfile
@@ -1,10 +1,12 @@
 # syntax=docker/dockerfile:1.4
-FROM node:16.17.0-bullseye-slim
+FROM node:16.17.0-bullseye-slim AS build
 ARG PLUGIN_VERSION
 RUN test -n "${PLUGIN_VERSION}"
 WORKDIR /app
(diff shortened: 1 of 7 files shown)


library/connect-web/v0.0.10/Dockerfile
```diff
@@ -1,10 +1,12 @@
...
-FROM node:16.17.0-bullseye-slim
+FROM node:16.17.0-bullseye-slim AS build
 ARG PLUGIN_VERSION
 RUN test -n "${PLUGIN_VERSION}"
 WORKDIR /app
 COPY --link package*.json .
 RUN npm ci
 
+FROM node:16.17.0-bullseye-slim
+COPY --link --from=build /app /app
```
pkwarren, line 10: Dug in some more and found that `/root/.npm` had cache which differed between builds.

mfridman (approved):

PR #69, opened 2022-09-08
allow specifying 'latest' as plugin version

pkwarren:
When building specific plugins, allow using 'latest' instead of a semantic version to only build the latest version of a given plugin. For example, `make PLUGINS="connect-web:latest"` will only build the latest version of connect-web.

### DIFF
--- a/internal/plugin/plugin.go
+++ b/internal/plugin/plugin.go
@@ -322,6 +322,9 @@ func parsePluginsEnvVar(pluginsEnv string) ([]includePlugin, error) {
 	return includes, nil
 }
 
+// getLatestPluginVersionsByName returns a map with keys set to plugin.Name and values set to the latest semver version for the plugin.
+// For example, if plugins contains buf.build/library/connect-web v0.1.1, v0.2.0, and v0.2.1,
+// the returned map will contain: {"buf.build/library/connect-web": "v0.2.1"}.
 func getLatestPluginVersionsByName(plugins []*Plugin) map[string]string {
 	latestVersions := make(map[string]string)
 	for _, plugin := range plugins {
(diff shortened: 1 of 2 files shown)


internal/plugin/plugin_test.go
```diff
@@ -36,6 +36,10 @@ func TestFilterByPluginsEnv(t *testing.T) {
 	assert.Equal(t, filterPluginsByPathPrefixes(t, plugins, "contrib/chrusty/jsonschema/"),
 		runFilterByPluginsEnv(t, plugins, "chrusty-jsonschema"))
 	assert.Equal(t, filterPluginsByPathPrefixes(t, plugins, "contrib/", "library/"), runFilterByPluginsEnv(t, plugins, ""))
+	latestConnectWeb := getLatestPluginVersionsByName(plugins)["buf.build/library/connect-web"]
+	require.NotEmpty(t, latestConnectWeb)
+	assert.Equal(t, filterPluginsByPathPrefixes(t, plugins, "library/connect-web/"+latestConnectWeb+"/"),
```
mfridman, line 41: Interesting, took me a second to grok what's going on here.

internal/plugin/plugin.go
```diff
@@ -321,16 +322,33 @@ func parsePluginsEnvVar(pluginsEnv string) ([]includePlugin, error) {
 	return includes, nil
 }
 
+func getLatestPluginVersionsByName(plugins []*Plugin) map[string]string {
```
mfridman, line 325: Nit, wouldn't mind a comment with an example of the final map.

mfridman (approved):

internal/plugin/plugin.go
pkwarren: Added.

Issue #73, opened 2022-09-11
Plugin request for Buf Schema Registry - protoc-gen-go-vtproto


mfridman:
## Mandatory

**Where is the source code for the plugin?**

github.com/planetscale/vtprotobuf/cmd/protoc-gen-go-vtproto

## Optional

**Does the plugin have a valid semver version?**

`v0.3.0` https://github.com/planetscale/vtprotobuf/releases/tag/v0.3.0

**Does the plugin have default options?**

N/A

**Does this plugin have runtime dependencies?**

N/A.

**Does the plugin have a dependency on another plugin?**

vtprotobuf is implemented as a helper plug-in that must be run alongside the upstream protoc-gen-go generato

**Do you think this plugin will be compatible with the BSR go module proxy or npm registry?**

N/A

pkwarren:
Fixed by #79.

bufdev:
Why did we add this? This isn't compatible with the go module proxy as currently implemented.

PR #79, opened 2022-09-15
Add vtproto plugin

oliversun09:
[Issue 73](https://github.com/bufbuild/plugins/issues/73)


library/vtproto/v0.3.0/buf.plugin.yaml
```diff
@@ -0,0 +1,13 @@
+version: v1
+name: buf.build/library/vtproto
+plugin_version: v0.3.0
+output_languages: [go]
+source_url: https://github.com/planetscale/vtprotobuf
+description: Generates optimized marshall & unmarshal code.
```
pkwarren, line 6:
```suggestion
description: generates optimized marshaling & unmarshaling Go code for ProtoBuf APIv2.
```

library/vtproto/v0.3.0/buf.plugin.yaml
```diff
@@ -0,0 +1,13 @@
...
+default_opts:
+  - paths=source_relative
+registry:
+  go:
+    deps:
+      - module: google.golang.org/protobuf
+        version: v1.28.1
```
mfridman, line 13: I presume the latest version will work, although I do wonder if we want to match this closer to the go.mod file?

https://github.com/planetscale/vtprotobuf/blob/main/go.mod#L8

cc @pkwarren thoughts?

library/vtproto/v0.3.0/buf.plugin.yaml
```diff
@@ -0,0 +1,13 @@
+version: v1
+name: buf.build/library/vtproto
```
pkwarren, line 2: We might want to name this `vtproto-go`? What do you think @mfridman / @smallsamantha? It might make it more discoverable and be similar to how we handle twirp.

library/vtproto/v0.3.0/buf.plugin.yaml
pkwarren: I think we should remove the registry section - it currently generates code which has to live in the same package as the generated protos. With the default features, it also requires grpc.

People using this will need to use `buf.gen.yaml` until we get a chance to patch plugins to work together.

pkwarren (changes requested):

library/vtproto/v0.3.0/buf.plugin.yaml
smallsamantha: yup i agree @pkwarren i think the consistency is a good call

mfridman: For the most part, I agree. My only concern is we have some standardized way to do this. 

Maybe in lieu of a plugin name that includes the language we suffix it?

library/vtproto/v0.3.0/buf.plugin.yaml
mfridman: Yep, let's remove registry config, esp. for plugins that output to the same package along with the base types.

library/vtproto/v0.3.0/buf.plugin.yaml
```diff
@@ -0,0 +1,13 @@
...
+plugin_version: v0.3.0
+output_languages: [go]
+source_url: https://github.com/planetscale/vtprotobuf
+description: Generates optimized marshall & unmarshal code.
+default_opts:
+  - paths=source_relative
```
pkwarren, line 8: I think we may want to set additional options (i.e. disable the experimental gRPC feature by default): https://github.com/planetscale/vtprotobuf/issues/25#issuecomment-1011055026

pkwarren (approved):

oliversun09:
I discarded this fix by mistake earlier

pkwarren (approved):

PR #81, opened 2022-09-16
Add BSR npm registry support for grpc-web and js

mfridman:
We may not want to merge this, but putting this up for a discussion.

### DIFF
--- a/library/protoc/v3.20.1/js/buf.plugin.yaml
+++ b/library/protoc/v3.20.1/js/buf.plugin.yaml
@@ -7,3 +7,9 @@ description: Official protobuf plugin for JavaScript.
 default_opts:
   - import_style=commonjs
   - binary
+registry:
+  npm:
+    deps:
+      # https://www.npmjs.com/package/google-protobuf
+      - package: "google-protobuf"
+        version: "^3.21.0"
--- a/library/grpc-web/v1.3.1/buf.plugin.yaml
+++ b/library/grpc-web/v1.3.1/buf.plugin.yaml
@@ -9,3 +9,9 @@ default_opts:
   - mode=grpcwebtext
 deps:
   - plugin: buf.build/library/js:v3.20.1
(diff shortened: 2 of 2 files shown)


mfridman:
If we do want to add support for `grpc-web` (which I think we do), what should be the default mode?

Right now we have `- mode=grpcwebtext`, but should this be `- mode= grpcweb `? 

@timostamm based on your experience, do you have a suggestion on the default mode for grpc-web that will satisfy _most_ users?

Note that we have the ability to set default options (used in buf.gen.yaml) _and_ a separate set of options for the BSR npm registry: `registry.opts`.

library/protoc/v3.20.1/js/buf.plugin.yaml
```diff
@@ -7,3 +7,9 @@ description: Official protobuf plugin for JavaScript.
 default_opts:
   - import_style=commonjs
   - binary
+registry:
+  npm:
+    deps:
+      # https://www.npmjs.com/package/google-protobuf
+      - package: "google-protobuf"
+        version: "^3.21.0"
```
pkwarren, line 15: Should this be 3.20.1 to match protoc version?

pkwarren (approved):

pkwarren (approved):

timostamm:
We don't support `grpcwebtext` in `connect-go`. The additional base64 encoded was originally used for Internet Explorer, because that was the only way to allow reading a streaming response via XMLHttpRequest. Internet Explorer is officially EOL since this summer, and as far as I know, all gRPC-web servers accept text _and_ binary. 

There might be other runtimes (react-native) that need it, so there might be a minority of users still needing it after all, but I think it is very reasonable to use `mode=grpcweb` by default.

pkwarren:
Before landing this, we should update the js v3.21.0 plugin introduced in #84 as well.

PR #112, opened 2022-09-26
only build/test/push changed plugins

pkwarren:
Update the ci workflow to determine which files changed since the last successful build, and only build/test/push the plugins which changed (instead of all of them).

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -22,7 +22,7 @@ jobs:
         fetch-depth: 0
     - name: Get changed files
       id: changed-files
-      uses: tj-actions/changed-files@v29.0.7
+      uses: tj-actions/changed-files@v31
       with:
         files: |
           contrib/**
--- a/.github/workflows/push_all.yml
+++ b/.github/workflows/push_all.yml
@@ -1,7 +1,5 @@
-name: ci
+name: push_all
 on:
-  push:
-    branches: [ "main" ]
   workflow_dispatch: {} # support manual runs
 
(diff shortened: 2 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -16,7 +16,38 @@ jobs:
     steps:
     - uses: actions/checkout@v3
       with:
-        fetch-depth: 1
+        fetch-depth: 0
+    - name: Get branch name
+      id: branch-name
+      uses: tj-actions/branch-names@v6
```
mfridman, line 22: Curious, why do we need this action? Not suggesting we change it, but don't we already have this from the github event payload: `github.ref_name` ?

https://docs.github.com/en/actions/learn-github-actions/contexts

.github/workflows/ci.yml
```diff
@@ -16,7 +16,38 @@ jobs:
...
+    - uses: nrwl/last-successful-commit-action@v1
+      id: last_successful_commit_push
+      with:
+        branch: ${{ steps.branch-name.outputs.current_branch }}
+        workflow_id: 'ci.yml'
+        github_token: ${{ secrets.GITHUB_TOKEN }}
+    - name: Get changed files
+      id: changed-files
+      uses: tj-actions/changed-files@v31
```
mfridman, line 31: More of a comment in passing, but 31 versions to get an action to detect changed-files? Seems a bit excessive, although maybe this evolved with the Actions platform and breaking changes were due to upstream changes 🤔

mfridman (approved):

PR #113, opened 2022-09-27
update the Go plugin source URL

pkwarren:


library/go/v1.28.0/buf.plugin.yaml
```diff
@@ -2,7 +2,7 @@ version: v1
 name: buf.build/library/go
 plugin_version: v1.28.0
 output_languages: [go]
-source_url: https://github.com/golang/protobuf
+source_url: https://github.com/protocolbuffers/protobuf-go
```
pkwarren, line 5: This was pointing to the previous Go v1 library.

mfridman (approved):

PR #114, opened 2022-09-28
improve the fetch_versions client

pkwarren:
Update the fetch_versions client to support fetching from the top-level directory (so contrib versions can be updated too). Add support for using GITHUB_TOKEN for authentication (to avoid rate limit issues). Do a replace across all files of the old version with the new version when creating the files for a new plugin.

### DIFF
--- a/.github/workflows/fetch_versions.yml
+++ b/.github/workflows/fetch_versions.yml
@@ -7,7 +7,6 @@ on:
 permissions:
   contents: write
   pull-requests: write
-  repository-projects: read
 
 jobs:
   fetch-versions:
--- a/.github/workflows/fetch_versions.yml
+++ b/.github/workflows/fetch_versions.yml
@@ -7,6 +7,7 @@ on:
 permissions:
   contents: write
   pull-requests: write
+  repository-projects: read
 
 jobs:
   fetch-versions:
@@ -15,8 +16,10 @@ jobs:
       - name: Checkout repository code
         uses: actions/checkout@v3
       - name: Fetch all versions
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: |
-          go run ./cmd/fetcher library
+          go run ./cmd/fetcher .
       - name: Create PR
         uses: peter-evans/create-pull-request@171dd555b9ab6b18fa02519fdfacbb8bf671e1b4
         with:
--- a/cmd/fetcher/main.go
+++ b/cmd/fetcher/main.go
@@ -43,7 +43,7 @@ func run(ctx context.Context, root string, depth int) error {
 	defer func() {
 		log.Printf("finished running in: %.2fs\n", time.Since(now).Seconds())
 	}()
-	configs, err := source.GatherConfigs(root, depth)
+	configs, err := source.GatherConfigs(root)
 	if err != nil {
 		return err
 	}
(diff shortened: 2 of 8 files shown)


.github/workflows/fetch_versions.yml
```diff
@@ -7,6 +7,7 @@ on:
 permissions:
   contents: write
   pull-requests: write
+  repository-projects: read
```
pkwarren, line 10: Since the fetchclient.go calls Repositories.ListTags, I thought this might be required?

.github/workflows/fetch_versions.yml
```diff
@@ -15,8 +16,10 @@ jobs:
       - name: Checkout repository code
         uses: actions/checkout@v3
       - name: Fetch all versions
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: |
-          go run ./cmd/fetcher library
+          go run ./cmd/fetcher .
```
pkwarren, line 22: This will now update all plugins in both library and contrib.

cmd/fetcher/main.go
```diff
@@ -226,12 +232,13 @@ func updateBufPluginFile(name string, newVersion string) error {
...
 }
 
-func copyFile(src, dest string) error {
+func copyFile(src, dest string, prevVersion, newVersion string) error {
 	data, err := os.ReadFile(src)
 	if err != nil {
 		return err
 	}
-	return os.WriteFile(dest, data, 0644)
+	replaced := strings.ReplaceAll(string(data), strings.TrimPrefix(prevVersion, "v"), strings.TrimPrefix(newVersion, "v"))
```
pkwarren, line 240: For now just does a naive replacement of the previous version with the new version across all files.

internal/fetchclient/fetchclient.go
```diff
@@ -33,10 +35,16 @@ type Client struct {
...
-}
-
-func newClient(client *http.Client) *Client {
+	var client *http.Client
+	if ghToken := os.Getenv("GITHUB_TOKEN"); ghToken != "" {
+		ctx := context.Background()
+		ts := oauth2.StaticTokenSource(
+			&oauth2.Token{AccessToken: ghToken},
+		)
+		client = oauth2.NewClient(ctx, ts)
```
pkwarren, line 44: I hit rate limits pretty easily locally so added support for using a PAT.

internal/source/source.go
```diff
@@ -47,36 +46,23 @@ func loadConfigFile(filename string) (_ *Config, retErr error) {
...
+					return filepath.SkipDir
 				}
 			}
-			if !found {
-				return fmt.Errorf("%w: %s", ErrSourceFileNotFound, path)
-			}
-			return nil
-		}
-		if d.Name() == "source.yaml" {
+		} else if d.Name() == "source.yaml" {
```
pkwarren, line 65: Relaxed checks in here to gather all of the `source.yaml` files at the top-level directory.

mfridman: Nice cleanup, now that I look at the existing code it was a bit of a mess.

cmd/fetcher/main.go
mfridman: Probably good enough too. This will do it for _all_ files, but that's probably fine.

mfridman (approved):

mfridman:
Nice cleanup, and the version search/replace will be a great help to reduce the manual bits we have to remember to update!

PR #116, opened 2022-09-29
add protoc-gen-validate cpp/go/java plugins

### LINKED ISSUE #71: Plugin request for Buf Schema Registry - protoc-gen-validate
## Mandatory

https://github.com/envoyproxy/protoc-gen-validate

## Optional

`v0.6.7`

https://github.com/envoyproxy/protoc-gen-validate/releases/tag/v0.6.7

**Does the plugin have default options?**

No, but the `lang` must be specified.. https://github.com/envoyproxy/protoc-gen-validate#parameters

N/A

Nope.

pkwarren:
Add protoc-gen-validate C++/Go/Java plugins.

Fixes #71.


library/protoc-gen-validate-cpp/v0.6.12/buf.plugin.yaml
```diff
@@ -0,0 +1,8 @@
+version: v1
+name: buf.build/library/protoc-gen-validate-cpp
```
mfridman, line 2: I don't think we need to call this `protoc-gen-validate-cpp`, maybe `validate-cpp` is enough?

library/validate-cpp/v0.6.12/Dockerfile
```diff
@@ -0,0 +1,11 @@
+# syntax=docker/dockerfile:1.4
+FROM golang:1.19.1-bullseye AS build
+WORKDIR /app
+COPY --link . /app
+RUN CGO_ENABLED=0 go install -ldflags "-s -w" -trimpath github.com/envoyproxy/protoc-gen-validate@v0.6.12
```
mfridman, line 5: This is because the module is still named like so?

https://github.com/bufbuild/protoc-gen-validate/blob/main/go.mod#L1

Do you know if there are any plans to change this, more a curiosity.

mfridman (approved):

mfridman:
I presume we also want to publish a _module_ called `protoc-gen-validate` .. or just `validate` to the BSR?

https://buf.build/envoyproxy/protoc-gen-validate

E.g., `buf.build/bufbuild/validate` maybe?

pkwarren:
> I presume we also want to publish a _module_ called `protoc-gen-validate` .. or just `validate` to the BSR?
> 
> https://buf.build/envoyproxy/protoc-gen-validate
> 
> E.g., `buf.build/bufbuild/validate` maybe?

I think we might want to keep it named `protoc-gen-validate` for discoverability, but either would work.

PR #130, opened 2022-10-04
only create a single branch for plugin updates

pkwarren:
Don't specify a `branch-suffix` so the default behavior is used (only one branch will be created and it will be updated as needed).


pkwarren:
See https://github.com/peter-evans/create-pull-request#action-behaviour.

mfridman (approved):

PR #145, opened 2022-10-10
update fetcher to retrieve latest version

pkwarren:
Maven versions can be returned out of order (the default sort is by score). Update the query to determine the latest semver version from the API (ignoring pre-release versions).

Additionally, fetch the latest patch version for each major+minor (we can update the utility in the future to use this to update to later patch versions of plugins we already ship).

### DIFF
--- a/internal/fetchclient/fetchclient.go
+++ b/internal/fetchclient/fetchclient.go
@@ -76,7 +76,11 @@ func (c *Client) fetch(ctx context.Context, config *source.Config) (string, erro
 	case config.Source.NPMRegistry != nil:
 		return c.fetchNPMRegistry(ctx, config.Source.NPMRegistry.Name)
 	case config.Source.Maven != nil:
-		return c.fetchMaven(ctx, config.Source.Maven.Group, config.Source.Maven.Name)
+		results, err := c.fetchMaven(ctx, config.Source.Maven.Group, config.Source.Maven.Name)
+		if err != nil {
+			return "", err
+		}
+		return results.latestVersion, nil
 	}
 	return "", fmt.Errorf("failed to match a source")
 }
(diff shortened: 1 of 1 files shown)


internal/fetchclient/fetchclient.go
```diff
@@ -172,22 +181,24 @@ func (c *Client) fetchMaven(ctx context.Context, group string, name string) (str
...
-		return "", err
+		return nil, err
 	}
 	defer response.Body.Close()
 
 	var data struct {
 		Response struct {
-			Docs []struct {
+			NumFound int `json:"numFound"`
+			Start    int `json:"start"`
```
pkwarren, line 200: Added these for future use (if we want to paginate over all versions). Right now scanning the first 20 results works ok.

internal/fetchclient/fetchclient.go
```diff
@@ -196,12 +207,36 @@ func (c *Client) fetchMaven(ctx context.Context, group string, name string) (str
...
+		version := doc.Version
+		if !strings.HasPrefix(version, "v") {
+			version = "v" + version
+		}
+		if !semver.IsValid(version) || semver.Prerelease(version) != "" {
+			continue
+		}
+		version = semver.Canonical(version)
+		if results.latestVersion == "" || semver.Compare(results.latestVersion, version) < 0 {
+			results.latestVersion = version
```
pkwarren, line 229: Keep track of the latest available version.

internal/fetchclient/fetchclient.go
```diff
@@ -196,12 +207,36 @@ func (c *Client) fetchMaven(ctx context.Context, group string, name string) (str
...
+		}
+		release := semver.MajorMinor(version)
+		if results.latestPatchVersions[release] == "" || semver.Compare(results.latestPatchVersions[release], version) < 0 {
+			results.latestPatchVersions[release] = version
```
pkwarren, line 233: Keep track of the latest version for a given major + minor release. We could use this in a follow up PR to update to the latest patch version of a plugin we're already shipping.

pkwarren:
This works around a bug in #131 where we aren't correctly detecting the latest version of the grpc-java plugin.

smallsamantha (approved):

mfridman:
> Maven versions can be returned out of order (the default sort is by score). Update the query to determine the latest semver version from the API (ignoring pre-release versions).

The changes look good to me, thanks for fixing this up. I'm curious, where does it say the default sort is by score? Afaics the docs say:

> Returns sorted list of all versions of an artifact.

Ref, the [second example](https://central.sonatype.org/search/rest-api-guide/).

PR #148, opened 2022-10-11
Restructure plugin naming and directories

mfridman:
This PR moves all plugin-related code to `/plugins` directory, where the first level is the owner name (owners are expected to exist on the BSR already):

```
plugins
├── apple
├── bufbuild
├── community
├── grpc
├── protocolbuffers
└── twitchtv
```

The `community` is a catch-all for plugins without a BSR owner. The convention is to name them:

```
{owner identifier}-{plugin name}
```

E.g., `planetscale-vtproto` or `mitchellh-go-json`. This is a best-effort, and the owner identifier is usually a GitHub owner.

### Questions

1. Should we add a language suffix if one does not exist? E.g., `planetscale-vtproto-go` instead of just `planetscale-vtproto` ?

No, let's keep it consistent with the upstream naming, stripping the `protoc-gen-` prefix and using the remainder as the plugin name

### TODO

- [x] ~We should move the group dir, in both grpc and protocolbuffers, into their own name even if they share a common base. That's an implementation detail I think, but let's chat.~ Yes, we're going with a consistent naming for plugins with common base images, at the expense of duplicating the code in the Dockerfile.
- [x] Fix broken tests, scripts..
- [x] ~Should `/protocolbuffers` instead be `/google` ?~ No, keep it as protocolbuffers for consistency with upstream repositories which themselves have been moved some time ago.


.github/workflows/ci.yml
```diff
@@ -29,12 +29,7 @@ jobs:
       with:
         base_sha: ${{ steps.last_successful_commit_push.outputs.commit_hash }}
         files: |
-          contrib/**
```
pkwarren, line 32: For builds on the main branch, we only want to look at changes to the plugins/* directory. We don't want to trigger a push of all plugins if tests or Makefile change.

.github/workflows/ci.yml
```diff
@@ -62,6 +55,7 @@ jobs:
       id: buildx
       uses: docker/setup-buildx-action@v2
       with:
+        driver: docker
```
pkwarren, line 58: This lets us cache build artifacts on local Docker daemon instead of more complex buildkit caching.

.github/workflows/pr.yml
```diff
@@ -8,6 +8,10 @@ on:
 # Ref: https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#permissions
 permissions: read-all
 
+concurrency:
+  group: ${{ github.ref }}
+  cancel-in-progress: true
```
pkwarren, line 13: We should only build the latest version of a PR branch and not queue up builds.

.github/workflows/pr.yml
```diff
@@ -25,8 +29,7 @@ jobs:
       uses: tj-actions/changed-files@v32
       with:
         files: |
-          contrib/**
-          library/**
+          plugins/**
```
pkwarren, line 32: We'll build/test everything if Makefile or tests/* change like before for PR builds only.

cmd/fetcher/main.go
```diff
@@ -48,18 +46,22 @@ func run(ctx context.Context, root string, depth int) error {
...
-		newVersion, err := client.Fetch(ctx, config)
-		if err != nil {
-			return err
+		newVersion := latestVersions[config.CacheKey()]
+		if newVersion == "" {
+			newVersion, err = client.Fetch(ctx, config)
+			if err != nil {
+				return err
+			}
+			latestVersions[config.CacheKey()] = newVersion
```
pkwarren, line 61: We've got a lot of plugins (grpc/protoc/etc.) which share the same source.yaml configuration. This allows us to optimize lookup of new versions to only fetch the latest one once.

cmd/fetcher/main.go
```diff
@@ -48,18 +46,22 @@ func run(ctx context.Context, root string, depth int) error {
...
 		}
 		// For now we ignore prerelease versions. But this may change in the future.
 		if semver.Prerelease(newVersion) != "" && !config.IncludePrerelease {
-			log.Printf("%s: skipping prerelease version: %s", config.Source.Name(), newVersion)
```
pkwarren, line 62: Removed a bunch of extra logging so Github actions log will only show the new discovered plugins.

internal/plugin/plugin.go
```diff
@@ -69,6 +70,13 @@ func Walk(dir string, f func(plugin *Plugin)) error {
 	}); err != nil {
 		return err
 	}
+	sort.SliceStable(unsorted, func(i, j int) bool {
+		p1, p2 := unsorted[i], unsorted[j]
+		if p1.Name != p2.Name {
+			return p1.Name < p2.Name
+		}
+		return semver.Compare(p1.Version, p2.Version) < 0
```
pkwarren, line 78: Prints out things in a nicer order by semver.

internal/plugin/plugin.go
```diff
@@ -91,16 +99,10 @@ func sortByDependencyOrder(original []*Plugin, pluginNames map[string]struct{})
...
 			foundDeps := true
 			for _, dep := range plugin.Deps {
-				name, _, ok := strings.Cut(dep.Plugin, ":")
+				_, _, ok := strings.Cut(dep.Plugin, ":")
 				if !ok {
 					return nil, fmt.Errorf("invalid plugin dependency: %s", dep)
 				}
-				if _, ok := pluginNames[name]; !ok {
-					// Plugin dependency defined elsewhere - this is ok.
-					// We only calculate dependency order for plugins found within the specified path.
```
pkwarren, line 100: Removed this so we check dependencies between plugins. For now, everything is expected to be contained in this repo.

plugins/apple/swift/v1.20.2/Dockerfile
```diff
@@ -6,7 +6,7 @@ RUN git clone --depth 1 --branch 1.20.2 https://github.com/apple/swift-protobuf
 WORKDIR /app/swift-protobuf
 RUN swift build -c release --static-swift-stdlib
 
-FROM debian:bullseye-20220912-slim
+FROM debian:bullseye-20221004-slim
```
pkwarren, line 9: Bumped everything to latest base images.

plugins/grpc/cpp/v1.49.1/Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+    if [ "${arch}" = "amd64" ]; then arch="x86_64"; fi; \
+    curl -fsSL -o /usr/local/bin/bazel https://github.com/bazelbuild/bazel/releases/download/4.2.2/bazel-4.2.2-linux-${arch} \
+ && chmod +x /usr/local/bin/bazel
+
+WORKDIR /build
+
+RUN git clone --depth 1 --branch v1.49.1 --recursive https://github.com/grpc/grpc
+WORKDIR /build/grpc
+RUN bazel build //src/compiler:grpc_plugin_support
+RUN bazel build //src/compiler:grpc_cpp_plugin
```
pkwarren, line 18: Updated to build grpc plugins using Bazel instead of cmake (recommended in grpc development docs). This also allows us to benefit from caching since we'll only build the `:grpc_plugin_support` rule once and then have to build a single binary in each of the plugins (uncached).

CONTRIBUTING.md
```diff
@@ -26,25 +26,25 @@ The build requires the following:
 
 ## Creating a new plugin
 
-Plugins are found in the top-level `library` or `contrib` directories, depending on their level of support by the Buf team.
+Plugins are found in the top-level `plugins` directory.
 
-To create a new plugin, add a new folder matching the last component of the plugin's name and its version (i.e. `mkdir -p library/plugin-name/vX.Y.Z`) and add a `buf.plugin.yaml` / `Dockerfile` / `.dockerignore` to the newly created directory.
-To verify the plugin builds properly, run `make PLUGINS="<plugin-name>"` to build the Docker image and `make test PLUGINS="<plugin-name>"` to verify code generation for the plugin using some basic APIs stored in `tests/testdata/images/`.
+To create a new plugin, add a new folder with the plugin's organization, name, and version (i.e. `mkdir -p plugins/<org>/<name>/<version>`) and add a `buf.plugin.yaml` / `Dockerfile` / `.dockerignore` to the newly created directory.
```
smallsamantha, line 31: minor thought: would we want to put something in the contribution guide about the community/other organizations distinction? or some disclaimer like "get approval from <someone> before adding new organization"

smallsamantha (approved):

CONTRIBUTING.md
pkwarren: Good point - we should probably add something here or to the .github/ISSUE_TEMPLATE/plugin-request-for-buf-schema-registry.md. I may add this in a follow up as the build takes a while to run on the PR branch (because it is rebuilding all plugins).

mfridman (commented):
Great job, thanks for taking this on 🎉 

Let's get this merged, but do not upload to the BSR as we have some backend changes to make.

Makefile
```diff
@@ -91,6 +51,13 @@ test: build
 		CACHE_ARGS="$${CACHE_ARGS} --cache-to type=registry,mode=max,ref=$(DOCKER_WRITE_CACHE_ORG)/plugins-$${PLUGIN_OWNER}-$${PLUGIN_NAME}:$${PLUGIN_VERSION}-buildcache"; \
 	fi; \
 	test -n "$${PLUGIN_NAME}" -a -n "$${PLUGIN_VERSION}" && \
+	$(DOCKER) $(DOCKER_BUILD_ARGS) \
+		$(DOCKER_BUILD_EXTRA_ARGS)$${CACHE_ARGS} \
+		--label build.buf.plugins.config.owner=$${PLUGIN_OWNER} \
```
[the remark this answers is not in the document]
pkwarren: The latter - just so we can look at an image in the OCI proxy and work backwards to the plugin.

PR #153, opened 2022-10-12
Remove js and go plugins incompatible with BSR

mfridman:
This PR removes go- and js-based plugins that are not compatible with BSR go proxy and npm registry.

- grpc/node
- grpc/web
- protocolbuffers/js
- twitchtv/twirp
- community/planetscale-vtprotobuf


pkwarren (changes requested):
We should also remove the associated files under tests/testdata.

pkwarren (approved):

mattrobenolt:
👋 from PlanetScale here.

What can I do to get our vtprotobuf to be compatible with BSR? Is there any information on what made it incompatible in the first place? I'm generally unfamiliar with the BSR and remote plugins, but would like to make this work.

mfridman:
Hey there, there are 2 distinct BSR features: **Remote Plugins** and **Remote Packages**

Briefly, **Remote Plugins** are hosted plugins executed in our secure sandbox and work with `buf generate`. This allows organizations and teams to have a consistent way to generate code. No more installing plugins locally or packaging "dev" containers.

**Remote Packages** takes this a step further. We implemented Go proxy and NPM registry, so folks can `go get ...` or `npm install ...` already generated and versioned code and don't have to deal with proto files, compilers, or plugins. Think of this as an opinionated way to package generate code exposed through tools most developers are familiar with.

E.g., `go get buf.build/gen/go/bufbuild/reflect/bufbuild/connect-go` fetches a regular Go package that is the product of the [bufbuild/reflect BSR Module](https://buf.build/bufbuild/reflect) against the [bufbuild/connect-go BSR Plugin](https://buf.build/bufbuild/connect-go).

---

To answer your question, adding `vtprotobuf` as a **Remote Plugin** is trivial. Just need to package it up as a docker container and anyone can reference this plugin as part of their `buf.gen.yaml`.

However, to make a plugin compatible with **Remote Packages** is another discussion altogether. Mainly because we expect all plugins to output code into separate packages, so we can build dependencies between plugins. The bufbuild/connect-go plugin is a great example. Although I suspect for `vtprotobuf` it'll be more challenging because the plugin outputs code into the same package as the base types.

I can add more context, but to make sure we're on the same page .. are you interested in adding this plugin as a Remote Plugin, Remote Package or both?

mattrobenolt:
Gotcha, that makes sense. So a remote plugin is what I'd be interested in. 

I already had https://buf.build/planetscale/plugins/vtproto which is being used by some folks, but there's a nice banner on this page that says "We are deactivating custom remote plugins. While your existing plugins will work in the interim, we recommend switching to remote plugins to avoid disruptions in service."

This request came to me through a mutual customer, so I'm just trying to piece together if I should say "you can keep using what we have" or if I need to do something in this repo now to be more correct.

The exact request I got is "I'd be perfect if you could convert this to a new buf plugin", and was assuming it's related to all of this.

If I'm misunderstanding, I'm happy to push back and ask them more questions. 

Thanks for the help!

mfridman:
> The exact request I got is "I'd be perfect if you could convert this to a new buf plugin", and was assuming it's related to all of this.

Yep, that's exactly it. I presume this is related to https://github.com/bufbuild/plugins/issues/251 .. so let's continue there. I can take this back to our team and report back on that issue.

The added advantage for plugins tracked within this repository is they are automatically kept up-to-date, a dependabot for plugins of sorts.

PR #157, opened 2022-10-13
add plugins SPDX license ID and install buf from main

cyinma:
fixes BSR-676

we need to install buf from main as the plugin license handling in buf CLI in not released yet

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -37,14 +37,14 @@ jobs:
       run: |
         echo '${{ toJSON(steps.changed-files.outputs) }}'
       shell: bash
-    - name: Install buf cli
-      run: |
-        go install github.com/bufbuild/buf/cmd/buf@main
     - name: Install Go
       uses: actions/setup-go@v3
       with:
         go-version: 1.19.x
         check-latest: true
+    - name: Install buf cli
+      run: |
+        go install github.com/bufbuild/buf/cmd/buf@main
     - name: Cache
       uses: actions/cache@v3
       with:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -37,7 +37,9 @@ jobs:
       run: |
         echo '${{ toJSON(steps.changed-files.outputs) }}'
       shell: bash
-    - uses: bufbuild/buf-setup-action@v1
+    - name: Install buf cli
+      run: |
+        go install github.com/bufbuild/buf/cmd/buf@main
     - name: Install Go
       uses: actions/setup-go@v3
       with:
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -41,14 +41,14 @@ jobs:
       run: |
         echo '${{ toJSON(steps.changed-files.outputs) }}'
       shell: bash
-    - name: Install buf cli
-      run: |
-        go install github.com/bufbuild/buf/cmd/buf@main
     - name: Install Go
       uses: actions/setup-go@v3
       with:
         go-version: 1.19.x
         check-latest: true
+    - name: Install buf cli
+      run: |
+        go install github.com/bufbuild/buf/cmd/buf@main
     - name: Cache
       uses: actions/cache@v3
       with:
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -41,7 +41,9 @@ jobs:
       run: |
         echo '${{ toJSON(steps.changed-files.outputs) }}'
       shell: bash
-    - uses: bufbuild/buf-setup-action@v1
+    - name: Install buf cli
+      run: |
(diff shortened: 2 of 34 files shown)


plugins/protocolbuffers/cpp/v21.7/buf.plugin.yaml
```diff
@@ -5,3 +5,4 @@ source_url: https://github.com/protocolbuffers/protobuf
 description: Official protobuf plugin for C++.
 output_languages:
   - cpp
+spdx_license_id: BSD-3-Clause
```
cyinma, line 8: this one is a bit tricky, as [its license](https://github.com/protocolbuffers/protobuf/blob/v21.7/LICENSE) is not recognized by github, so I put it into [licensecheck](https://github.com/google/licensecheck) and it suggested it is [BSD-3-Clause](https://spdx.org/licenses/BSD-3-Clause.html). it looks really similar but it will be great if someone can also have a look

.github/workflows/ci.yml
```diff
@@ -37,7 +37,9 @@ jobs:
       run: |
         echo '${{ toJSON(steps.changed-files.outputs) }}'
       shell: bash
-    - uses: bufbuild/buf-setup-action@v1
+    - name: Install buf cli
+      run: |
+        go install github.com/bufbuild/buf/cmd/buf@main
```
pkwarren, line 42: I think this should come after 'Install Go' step.

.github/workflows/pr.yml
```diff
@@ -41,7 +41,9 @@ jobs:
       run: |
         echo '${{ toJSON(steps.changed-files.outputs) }}'
       shell: bash
-    - uses: bufbuild/buf-setup-action@v1
+    - name: Install buf cli
```
pkwarren, line 44: Same here - should happen after 'Install Go'.

plugins/protocolbuffers/cpp/v21.7/buf.plugin.yaml
pkwarren: That looks right to me - https://github.com/protocolbuffers/protobuf/commit/24bf56fb3a2fab42d355b15de11782c3144b9e80 mentions a change from Apache 2 to "New BSD", which corresponds to "[BSD 3-Clause "New" or "Revised" License](https://spdx.org/licenses/BSD-3-Clause.html).

pkwarren:
You will also need to bump the version in the `go.mod` file: https://github.com/bufbuild/plugins/blob/af4535b01335c9ecd1ce7c770dae500476f2ed78/go.mod#L6. We use this in unit tests to sanity check that the buf.plugin.yaml file contents can be parsed.

mfridman (commented):
Just a general question, is `spdx_license_id` a mandatory field in the buf.plugin.yaml file?

Should we always enforce a plugin license, or is that too restrictive?

pkwarren:
> Should we always enforce a plugin license, or is that too restrictive?

We could enforce that in this repository (and write a unit test to ensure it). In previous code reviews we've preferred not to enforce these type of restrictions in the CLI and add them to the BSR if needed.

cyinma:
> Just a general question, is `spdx_license_id` a mandatory field in the buf.plugin.yaml file?
> 
> Should we always enforce a plugin license, or is that too restrictive?

its not, maybe we can do some kind of checking in this repository to make sure every plugins we push here included a license (which is expected I believe), but doing that in CLI to enforce this in general sounds too restrictive

mfridman:
That's kind of where I was going with this, we could extend this check to enforce some fields? Agreed we should keep it out of the CLI.

https://github.com/bufbuild/plugins/blob/main/tests/plugins_test.go#L122-L123

tests/plugins_test.go
```diff
@@ -119,8 +119,11 @@ func TestBufPluginConfig(t *testing.T) {
 	for _, p := range plugins {
 		yamlBytes, err := os.ReadFile(p.Path)
 		require.NoError(t, err)
-		_, err = bufpluginconfig.GetConfigForData(context.Background(), yamlBytes)
+		config, err := bufpluginconfig.GetConfigForData(context.Background(), yamlBytes)
 		assert.NoErrorf(t, err, "invalid plugin config: %q", p.Path)
+		assert.NotEmpty(t, config.Name)
+		assert.NotEmpty(t, config.PluginVersion)
+		assert.NotEmpty(t, config.SPDXLicenseID)
```
cyinma, line 126: these are the values that I think we mush have, not sure if this is all of them?

mfridman: I think this is a good enough starting point. 

Nit, we should probably change assert to require for the `config, err` case, otherwise this will cause a panic?

mfridman (approved):

pkwarren (approved):

PR #160, opened 2022-10-17
update descriptions for plugins

pkwarren:
Make a pass to update the descriptions of plugins to better describe what they do.

### DIFF
--- a/plugins/bufbuild/connect-go/v1.0.0/buf.plugin.yaml
+++ b/plugins/bufbuild/connect-go/v1.0.0/buf.plugin.yaml
@@ -2,7 +2,7 @@ version: v1
 name: buf.build/bufbuild/connect-go
 plugin_version: v1.0.0
 source_url: https://github.com/bufbuild/connect-go
-description: Client and server code generator for connect-go. Compatible with the gRPC, gRPC-Web, and Connect RPC protocols.
+description: Generates client and server stubs for connect-go. Compatible with the gRPC, gRPC-Web, and Connect RPC protocols.
 deps:
   - plugin: buf.build/protocolbuffers/go:v1.28.1
 default_opts:
--- a/plugins/bufbuild/connect-go/v1.0.0/buf.plugin.yaml
+++ b/plugins/bufbuild/connect-go/v1.0.0/buf.plugin.yaml
@@ -2,7 +2,7 @@ version: v1
 name: buf.build/bufbuild/connect-go
 plugin_version: v1.0.0
 source_url: github.com/bufbuild/connect-go
-description: A better gRPC.
+description: Type-safe client and server code generator for connect-go.
 deps:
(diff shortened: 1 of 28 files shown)


mfridman (approved):

timostamm (approved):
Both connect-web related plugin descriptions work for me 👍

plugins/bufbuild/connect-web/v0.2.1/buf.plugin.yaml
```diff
@@ -2,7 +2,7 @@ version: v1
 name: buf.build/bufbuild/connect-web
 plugin_version: v0.2.1
 source_url: https://github.com/bufbuild/connect-web
```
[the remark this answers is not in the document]
pkwarren: Understood - the same could be said for almost all of the URLs. I think having a link to the top-level project should be ok - otherwise using something like the NPM package URL would be good as it is unlikely to change. This gets more complicated for plugins like https://github.com/grpc/grpc/blob/master/src/compiler/cpp_generator.cc / https://github.com/apple/swift-protobuf/tree/main/Sources/protoc-gen-swift / etc.

plugins/community/chrusty-jsonschema/v1.3.9/buf.plugin.yaml
```diff
@@ -2,5 +2,5 @@ version: v1
 name: buf.build/community/chrusty-jsonschema
 plugin_version: v1.3.9
 source_url: https://github.com/chrusty/protoc-gen-jsonschema
-description: Converts protobuf definitions into JSONSchemas, which can be used to dynamically validate JSON messages.
+description: Converts protobuf definitions into JSON Schemas, which can be used to dynamically validate JSON messages.
```
akshayjshah, line 5: Should "Schema" be lower-case?

plugins/bufbuild/connect-go/v1.0.0/buf.plugin.yaml
```diff
@@ -1,8 +1,8 @@
 version: v1
 name: buf.build/bufbuild/connect-go
 plugin_version: v1.0.0
-source_url: github.com/bufbuild/connect-go
-description: A better gRPC.
+source_url: https://github.com/bufbuild/connect-go
+description: Type-safe client and server code generator for connect-go.
```
akshayjshah, line 5: lgtm :tada:

If there's space for more text, I'd love to add "Compatible with the gRPC, gRPC-Web, and Connect RPC protocols."

plugins/community/chrusty-jsonschema/v1.3.9/buf.plugin.yaml
pkwarren: I was following examples here: https://json-schema.org/. The GitHub repo description also uses "JSON-Schema".

PR #161, opened 2022-10-17
Authenticate with GitHub App generated tokens

ElliotMJackson:
`GITHUB_TOKEN` cannot trigger action run as this could result in an infinite loop of runs and commits. 

> A GitHub App can be created for the sole purpose of generating tokens for use with GitHub actions. These tokens can be used in place of GITHUB_TOKEN or a [Personal Access Token (PAT)](https://docs.github.com/en/github/authenticating-to-github/creating-a-personal-access-token). GitHub App generated tokens are more secure than using a PAT because GitHub App access permissions can be set with finer granularity and are scoped to only repositories where the App is installed.

[SOURCE](https://github.com/peter-evans/create-pull-request/blob/main/docs/concepts-guidelines.md#authenticating-with-github-app-generated-tokens)

This change will mean the token used to raise the PR is different than those used in the actions and therefor enable the build to run without human intervention. 

TODO:
- [x] create github app 
- [x] configure app id
- [x] test with mock run


mfridman:
With recent GitHub Actions updates last month, is this still necessary?

https://github.blog/changelog/2022-09-08-github-actions-use-github_token-with-workflow_dispatch-and-repository_dispatch/

ElliotMJackson:
on a manually triggered run, sure - GITHUB_TOKEN will continue to work... this will enable the `cron` to work. But also, generally speaking, this is "the way to do it from now on"

mfridman:
> on a manually triggered run, sure - GITHUB_TOKEN will continue to work... this will enable the `cron` to work. But also, generally speaking, this is "the way to do it from now on"

Gotcha. 👍

mfridman (approved):

PR #173, opened 2022-10-24
BSR-691/fill license url in plugins

oliversun09:
- update dependencies
- Add license_url to plugins' `buf.plugin.yaml`


plugins/apple/swift/v1.20.2/buf.plugin.yaml
```diff
@@ -6,3 +6,4 @@ description: Base types for Swift. Generates message and enum types.
 output_languages:
   - swift
 spdx_license_id: Apache-2.0
+license_url: https://github.com/apple/swift-protobuf/blob/1.20.2/LICENSE.txt
```
pkwarren, line 9: This is using a standard apache 2 license - I'd prefer we not add it to buf.plugin.yaml and return a URL from the spdx_license_id.

plugins/protocolbuffers/cpp/v21.7/buf.plugin.yaml
```diff
@@ -6,3 +6,4 @@ description: Base types for C++. Generates message and enum types.
 output_languages:
   - cpp
 spdx_license_id: BSD-3-Clause
+license_url: https://github.com/protocolbuffers/protobuf/blob/v21.7/LICENSE
```
pkwarren, line 9: When I looked earlier, this is the only plugin with a custom license (BSD-3-Clause plus an added statement at the end regarding generated code.

plugins/apple/swift/v1.20.2/buf.plugin.yaml
oliversun09: I was going to agree with you. However, I just took a look at some of the apache 2 licenses from the links I added, and I saw three different files:
https://github.com/apple/swift-protobuf/blob/1.20.2/LICENSE.txt has 211 lines
https://github.com/bufbuild/connect-go/blob/v1.1.0/LICENSE has 201 lines
https://github.com/grpc/grpc/blob/v1.50.0/LICENSE has 610 lines
Also, it seems [this line](https://github.com/apple/swift-protobuf/blob/1.20.2/LICENSE.txt#L189) of the license is meant to be replaced to include the owner's name (although this particular plugin's maintainers did not do it), and we can expect different plugs' licenses to be different on this line.

mfridman: For consistency, shouldn't we use the license from the upstream source? E.g., this license also includes an exception..

https://github.com/apple/swift-protobuf/blob/1.20.2/LICENSE.txt#L205-L211

Apologies, I might have missed a few details. If `license_url` is omitted where does the license come from?

pkwarren: Ahh I see - the Apple swift license has a runtime exception at the bottom and the gRPC license lists three licenses - Apache 2, BSD 3-Clause, and MPL v2. This might mean that we need to extend the data model to accommodate multiple licenses.

@mfridman I was optimistic we might be able to refer to https://github.com/spdx/license-list-data/blob/v3.18/json/licenses.json (reference field) for standard licenses.

mfridman: I think that would have been my first instinct as well, to just use the well-known license from spdx.

However, this means we might misrepresent a license. So using the upstream source seems like the safest solution for us.

Re. multiple licenses, this is a tricky one. I err on the side of caution, again, and deferring to the upstream. GitHub itself doesn't support multiple licenses because the underlying gem package doesn't support it. https://github.com/licensee/licensee/issues/441

My suggestion here is to:

- use a single spdx license identifier, probably originating from the upstream source
- use the URL from the source code itself, otherwise default back to the spdx list

Happy to chat about this, I think it's worth putting some thought here.

PR #174, opened 2022-10-24
User docker login action, do not login to registry within PRs

tboerger:
Logging into the registry fails for pull requests, I guess they are protected secrets. Beside that PRs are using a registry service, so no need to authenticate to DockerHub or ghcr.io.


pkwarren:
@tboerger - Thanks for the PR but I'd like to try a different approach where we still use the Docker registry credentials for PRs with access (while fixing builds for PRs raised from forks). In the past we've run into rate limiting issues which is why we added it.

tboerger:
But also in that case you could use the login action instead of the shell snippet :)

PR #177, opened 2022-10-27
conditionally log in to dockerhub for PRs

pkwarren:
PRs raised from forks of the repository don't have permission to use secrets for Docker Hub. Update the PR workflow to make logging in to Docker Hub conditional (the build should still work but may be subject to rate limiting).

Update to use docker-login-action consistently in other places to log into the GitHub container registry.

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -14,6 +14,8 @@ concurrency:
 
 jobs:
   pr:
+    env:
+      dockerhub_username: ${{ secrets.DOCKERHUB_USERNAME }}
     runs-on: ubuntu-latest
     services:
       registry:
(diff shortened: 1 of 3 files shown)


.github/workflows/pr.yml
```diff
@@ -14,6 +14,8 @@ concurrency:
 
 jobs:
   pr:
+    env:
+      dockerhub_username: ${{ secrets.DOCKERHUB_USERNAME }}
```
pkwarren, line 18: Trying the pattern documented here https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-using-secrets to see if it allows us to conditionally execute a step if the secret is available.

.github/workflows/pr.yml
```diff
@@ -63,13 +65,18 @@ jobs:
       with:
         driver: docker
         driver-opts: network=host
-    - name: Login to Docker Hub
+    - if: ${{ env.dockerhub_username != '' }}
```
mfridman, line 68: I presume this is because you can't use github secrets in conditionals?

I think yes, but just double-checking the reason for setting this as an env variable.

mfridman (approved):
lgtm. Just one question

PR #185, opened 2022-11-01
Remove grpc/node from dependabot tracked Dockerfiles

rubensf:
We removed this Dockerfile in #153, so the dependabot fails whenever it tries to update this image. (eg
[here](https://github.com/bufbuild/plugins/network/updates/496922854)


mfridman:
We're almost certainly going to re-add this Dockerfile, I suggest we close this for now. And if we don't I'll make sure to reopen and merge this.

PR #191, opened 2022-11-02
support pagination over Maven query results

pkwarren:


internal/fetchclient/fetchclient.go
```diff
@@ -173,10 +177,69 @@ func (c *Client) fetchMaven(ctx context.Context, group string, name string) (*ma
 	if err != nil {
 		return nil, err
 	}
+	results := &mavenResults{
+		latestPatchVersions: make(map[string]string),
+	}
+	numFound := math.MaxInt
+	for start := 0; start < numFound; start += mavenPageSize {
+		response, err := c.fetchMavenPage(ctx, targetURL, group, name, start)
```
mfridman, line 185: Are we at risk of getting rate-limited with this?

Iirc the search results return the latest ~20 items, I might be misremembering but what problem do we solve by iterating over all versions instead of just the latest 20?

pkwarren: They come back in a non-deterministic order (seen in previous issues). Paginating over the results seems like the best option to try.

If we get rate limited we can dial this back, but as of right now we make a total of 5 API calls to them.

mfridman: Ah, for some reason I keep thinking it's sorted (we must have talked about this before)

mfridman (approved):

PR #192, opened 2022-11-03
add protocolbuffers/js plugin

pkwarren:


pkwarren:
This is on hold for now until we add metadata for common-js vs. ESM modules.

mfridman (approved):

PR #193, opened 2022-11-03
Plugin updates

bufdev:
In line with separating remote packages and remote plugins as two separate products.

- Remove all default options. Although I think we've talked about this at times, this behavior was surprising to me and @elliotmjackson, and isn't surfaced in the UI. Any options required for remote packages is now moved to be remote-package-specific. Defaults like this will likely move to recipes in the future.
- Delete the multiple plugins for protoc-gen-validate. protoc-gen-validate doesn't work with remote packages as of now anyways, and this doesn't map cleanly to the local plugin experience.


pkwarren:
Fixed by #198 and #206.

PR #194, opened 2022-11-04
install buf CLI as part of fetch versions

pkwarren:
We use the buf CLI in order to test code generation of newer plugins. Update the workflow to install Go, cache Go modules, and install the latest buf CLI.

### DIFF
--- a/.github/workflows/fetch_versions.yml
+++ b/.github/workflows/fetch_versions.yml
@@ -22,6 +22,20 @@ jobs:
         uses: actions/checkout@v3
         with:
           token: ${{ steps.generate_token.outputs.token }}
+      - name: Install Go
+        uses: actions/setup-go@v3
+        with:
+          go-version: 1.19.x
+          check-latest: true
+      - name: Cache
+        uses: actions/cache@v3
+        with:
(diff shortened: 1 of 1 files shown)


robbertvanginkel (approved):

.github/workflows/fetch_versions.yml
```diff
@@ -22,6 +22,20 @@ jobs:
...
+          check-latest: true
+      - name: Cache
+        uses: actions/cache@v3
+        with:
+          path: ~/go/pkg/mod
+          key: ${{ runner.os }}-plugins-fetch-versions-${{ hashFiles('**/go.sum') }}
+          restore-keys: ${{ runner.os }}-plugins-ci
+      - name: Install buf cli
+        run: |
+          go install github.com/bufbuild/buf/cmd/buf@main
```
robbertvanginkel, line 38: Do we need main? Or should we use https://github.com/bufbuild/buf-setup-action

pkwarren: We need main because it has support for some newer fields in `buf.plugin.yaml` (like licenses, etc.) that haven't made it into a release yet.

PR #206, opened 2022-11-07
remove default_opts from plugins

pkwarren:
Remove default_opts from all plugins (except for protoc-gen-validate v0.6.13 - will need to determine how best to make the change there).

This changes generated code for quite a few plugins since we are no longer generating with default opts.


internal/plugin/plugin.go
```diff
@@ -11,19 +12,18 @@ import (
...
-	Path        string       `yaml:"-"`
-	Relpath     string       `yaml:"-"`
-	Name        string       `yaml:"name"`
-	Version     string       `yaml:"plugin_version"`
-	Deps        []Dependency `yaml:"deps,omitempty"`
-	DefaultOpts []string     `yaml:"default_opts,omitempty"`
+	Path    string `yaml:"-"`
+	Relpath string `yaml:"-"`
+	// Parsed external yaml config
+	bufpluginconfig.ExternalConfig `yaml:"-"`
```
pkwarren, line 26: Instead of duplicating reading from yaml I'm just embedding the struct we use in buf's CLI here (this is just for tests/builds).

internal/plugin/plugin.go
```diff
@@ -75,7 +75,7 @@ func Walk(dir string, f func(plugin *Plugin)) error {
 		if p1.Name != p2.Name {
 			return p1.Name < p2.Name
 		}
-		return semver.Compare(p1.Version, p2.Version) < 0
+		return semver.Compare(p1.PluginVersion, p2.PluginVersion) < 0
```
pkwarren, line 78: The ExternalConfig Version is set to `v1` everywhere - we want to use the `PluginVersion` everywhere we used `Version` before.

tests/plugins_test.go
```diff
@@ -143,9 +143,13 @@ func createBufGenYaml(t *testing.T, basedir string, plugin *plugin.Plugin) error
 	defer func() {
 		require.NoError(t, bufGenYaml.Close())
 	}()
+	opts := plugin.ExternalConfig.Registry.Opts
```
pkwarren, line 146: Changed the tests to prefer running with registry opts (so we continue to get test coverage there - including on our patches).

tests/plugins_test.go
```diff
@@ -143,9 +143,13 @@ func createBufGenYaml(t *testing.T, basedir string, plugin *plugin.Plugin) error
...
+	if len(opts) == 0 {
+		opts = plugin.DefaultOpts
```
pkwarren, line 148: This is only needed for validate v0.6.13 plugins.

tests/testdata/buf.build/grpc/go/v1.2.0/petapis/plugin.sum
```diff
@@ -1 +1 @@
-h1:tOsD3cb4cNwALEOJpfypEprBVC7eVac8wEuesWizT8c=
+h1:3fylldA9SEwekfPvkDtpVP+PntCFSjLK5iT3Xl+3Pu8=
```
pkwarren, line 1: These are changing because we're now running tests with the `separate_package=true` option.

pkwarren:
This change along with #198 should fix the issues reported in #193.

tests/plugins_test.go
mfridman: Let's remove `v0.6.13` entirely and start publishing from `v0.9.0` onwards

pkwarren: Done (although we'll need a BSR migration to remove existing published plugins).

internal/plugin/plugin.go
```diff
@@ -126,21 +126,20 @@ func sortByDependencyOrder(original []*Plugin) ([]*Plugin, error) {
...
-	if err := yaml.NewDecoder(f).Decode(&plugin); err != nil {
-		return nil, err
-	}
 	plugin.Path = filepath.ToSlash(path)
 	plugin.Relpath, err = filepath.Rel(basedir, path)
 	if err != nil {
 		return nil, err
 	}
 	plugin.Relpath = filepath.ToSlash(plugin.Relpath)
+	if err := yaml.NewDecoder(bytes.NewReader(contents)).Decode(&plugin.ExternalConfig); err != nil {
```
mfridman, line 140: Do we want this to be a strict, or do we have checks earlier in the process that make sure the files are already valid? 

I.e., maybe we can use the config parser from bufbuild/buf?

tests/plugins_test.go
mfridman: Not sure I follow this change. Shouldn't we rely on the default plugin behavior?

pkwarren: We've always run with `default_opts` - it feels good to run with some sort of options if we specify them in buf.plugin.yaml to get more coverage of additional features.

Alternatively we could have two executions of a plugin - one with opts and one without, but that would be 2x the plugin.sum files we check in to this repo.

internal/plugin/plugin.go
pkwarren: Switched to using encoding.UnmarshalJSONOrYAMLStrict. What I really want here is the external representation of the config, not the Plugin interface (which would require doing things like mapping to/from map/list options for tests).

tests/plugins_test.go
mfridman: Based on the plugins we have today, do all of them work out-of-the-box without supplying options?

I guess my wish is _every_ plugin in existence shipped with sane options without requiring mandatory ones. IMO that would be a more accurate reflection of its behavior, esp. for `buf generate` users.

pkwarren: All of them work out of the box without options. By including options however, we're getting more coverage (including on things like patched plugins like grpc-go).

internal/plugin/plugin.go
mfridman: Me too.

mfridman (approved):

Makefile
```diff
@@ -44,13 +44,14 @@ test: build
...
-	$(DOCKER) $(DOCKER_BUILD_ARGS) \
+	cd $(<D) && $(DOCKER) $(DOCKER_BUILD_ARGS) \
 		$(DOCKER_BUILD_EXTRA_ARGS) \
 		--label build.buf.plugins.config.owner=$${PLUGIN_OWNER} \
 		--label build.buf.plugins.config.name=$${PLUGIN_NAME} \
 		--label build.buf.plugins.config.version=$${PLUGIN_VERSION} \
 		-t $(DOCKER_ORG)/plugins-$${PLUGIN_OWNER}-$${PLUGIN_NAME}:$${PLUGIN_VERSION} \
-		$(<D)
+		-f - \
+		. < Dockerfile
```
pkwarren, line 54: Works around https://github.com/moby/buildkit/issues/1368

mfridman: Could we `touch` the file to change its timestamp and add a comment on why?

mfridman (approved):

Makefile
```diff
@@ -44,14 +44,14 @@ test: build
 	if [[ "$(DOCKER_ORG)" = "ghcr.io/bufbuild" ]]; then \
 		$(DOCKER) pull $(DOCKER_ORG)/plugins-$${PLUGIN_OWNER}-$${PLUGIN_NAME}:$${PLUGIN_VERSION} || :; \
 	fi; \
-	cd $(<D) && $(DOCKER) $(DOCKER_BUILD_ARGS) \
+	touch $<; \
```
mfridman, line 47: Ah, of course this is the name of the first prerequisite. long live `$<`

mfridman (approved):
I'm okay to merge this as-is to unblock the removal of default-opts.

But let's chat about our long-term solution for these protocolbuffers and grpc plugins. Right now there is a bit of inconsistency in how we structure the files (inline vs copied in), and imo they should all be the same.

We should also explore a common `base-protoc-v21.7` image or artifact, but ideally after we have come up with a solution we're happy to address https://github.com/moby/buildkit/issues/1368

mfridman:
Great job for coming up with a creative solution @pkwarren! Thanks for seeing this one through.

PR #221, opened 2022-11-18
update fetcher to allow non-zero exit code from diff

pkwarren:
When we generate diffs from the previous generated code to the new generated code, we should allow the command to return an exit code of 1. Diff will return 1 when there are diffs, 0 when there are no diffs, and anything else if there are other errors.


pkwarren:
This will fix the error seen here: https://github.com/bufbuild/plugins/actions/runs/3493014724/jobs/5847372558

saquibmian:
Are we just logging the diffs for informational purposes, then?

saquibmian (approved):

PR #230, opened 2022-11-23
create Go CLI to automate plugin releases

pkwarren:
Create a Go CLI (cmd/release/main.go) which automates the creation of plugin releases. A release consists of the following:

* plugin-releases.json - Contains metadata about all plugins (including name, version, digests).
* plugin-releases.json.minisign - Minisign signature of plugin-releases.json.
* <owner>-<name>-<version>.zip - One or more zip files containing a buf.plugin.yaml and image.tar (created with docker save).

The CLI is configured to run once daily. It detects any changes since the latest release and will upload only the added/changed plugins.

There are two supported options to the CLI:

* `-minisign-private-key <key>`: Path to minisign private key. If omitted, the plugin-releases.json file will not be signed.
* `-dry-run`: If true, the release will be created locally and saved to a temporary directory on disk, but no changes will be made to GitHub.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -60,6 +60,7 @@ jobs:
         username: ${{ secrets.DOCKERHUB_USERNAME }}
         password: ${{ secrets.DOCKERHUB_TOKEN }}
     - name: Login to GitHub Container Registry
+      if: github.repository == 'bufbuild/plugins'
       uses: docker/login-action@v2
       with:
         registry: ghcr.io
@@ -78,6 +79,7 @@ jobs:
       shell: bash
       run: make test DOCKER_ORG="ghcr.io/bufbuild"
     - name: Push
+      if: github.repository == 'bufbuild/plugins'
       shell: bash
       env:
         ALL_MODIFIED_FILES: ${{ steps.changed-files.outputs.all_modified_files }}
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -15,7 +15,6 @@ concurrency:
 jobs:
   pr:
     env:
-      dockerhub_username: ${{ secrets.DOCKERHUB_USERNAME }}
       BUILDKIT_PROGRESS: plain
     runs-on: ubuntu-latest
     steps:
@@ -56,8 +55,8 @@ jobs:
       with:
         driver: docker
         driver-opts: network=host
-    - if: ${{ env.dockerhub_username != '' }}
-      name: Login to Docker Hub
+    - name: Login to Docker Hub
+      if: github.repository == 'bufbuild/plugins'
       uses: docker/login-action@v2
       with:
         username: ${{ secrets.DOCKERHUB_USERNAME }}
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 .build/
 .idea/
 .vscode/
+minisign.*
 node_modules/
 tests/testdata/**/buf.gen.yaml
 tests/testdata/**/gen/
--- a/cmd/release/main.go
+++ b/cmd/release/main.go
@@ -1,0 +1,583 @@
+package main
+
+import (
+	"archive/zip"
+	"context"
+	"crypto/sha256"
+	"encoding/hex"
+	"encoding/json"
+	"errors"
+	"flag"
+	"fmt"
+	"io"
+	"io/fs"
+	"log"
+	"net/http"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"sort"
+	"strconv"
+	"strings"
+	"time"
+
+	"aead.dev/minisign"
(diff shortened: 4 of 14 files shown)


.github/workflows/ci.yml
```diff
@@ -60,6 +60,7 @@ jobs:
         username: ${{ secrets.DOCKERHUB_USERNAME }}
         password: ${{ secrets.DOCKERHUB_TOKEN }}
     - name: Login to GitHub Container Registry
+      if: github.repository == 'bufbuild/plugins'
```
pkwarren, line 63: Made steps of existing workflows that can't be run on forks behind a conditional.

.github/workflows/pr.yml
```diff
@@ -56,8 +55,8 @@ jobs:
       with:
         driver: docker
         driver-opts: network=host
-    - if: ${{ env.dockerhub_username != '' }}
-      name: Login to Docker Hub
+    - name: Login to Docker Hub
+      if: github.repository == 'bufbuild/plugins'
```
pkwarren, line 59: Instead of testing if we can get a secret, only run on our original repo.

.gitignore
```diff
@@ -1,6 +1,7 @@
 .build/
 .idea/
 .vscode/
+minisign.*
```
pkwarren, line 4: I created a key locally for testing and didn't want to check it in (not any security risk - this was just for E2E testing).

internal/plugin/plugin.go
```diff
@@ -81,7 +81,9 @@ func Walk(dir string, f func(plugin *Plugin)) error {
 		return err
 	}
 	for _, p := range sorted {
-		f(p)
+		if err := f(p); err != nil {
+			return err
```
pkwarren, line 85: I had a fair amount of logic while walking plugins and wanted to fail fast, so updated this method signature to allow it.

cmd/release/main.go
```diff
@@ -0,0 +1,583 @@
...
+	"github.com/google/go-github/v48/github"
+	"github.com/hashicorp/go-retryablehttp"
+	"golang.org/x/mod/semver"
+	"golang.org/x/oauth2"
+)
+
+const (
+	githubOwner = "bufbuild"
+	// This is separate from githubOwner for testing releases (can point at personal fork)
+	githubReleaseOwner   = githubOwner // "pkwarren"
```
pkwarren, line 36: This was useful to test things against my fork.

cmd/release/main.go
```diff
@@ -0,0 +1,583 @@
...
+	flag.StringVar(&minisignPrivateKey, "minisign-private-key", "", "path to minisign private key file")
+	flag.Parse()
+
+	if len(flag.Args()) != 1 {
+		_, _ = fmt.Fprintf(os.Stderr, "usage: %s [-dry-run] [-minisign-private-key <file>] <directory>\n", os.Args)
+		os.Exit(2)
+	}
+	root := flag.Args()[0]
+	if err := run(root, minisignPrivateKey, dryRun); err != nil {
+		log.Fatalln(err.Error())
```
pkwarren, line 85: This is the only fatal logger - this ensures we clean things up (like temporary directories) on error.

cmd/release/main.go
```diff
@@ -0,0 +1,583 @@
...
+}
+
+func run(root string, minisignPrivateKey string, dryRun bool) error {
+	// Create temporary directory
+	tmpDir, err := os.MkdirTemp("", "plugins-release")
+	if err != nil {
+		return fmt.Errorf("failed to create temporary directory: %w", err)
+	}
+	log.Printf("created tmp dir: %s", tmpDir)
+	if !dryRun {
```
pkwarren, line 96: With `--dry-run`, it is useful to inspect what would be uploaded.

.golangci.yml
```diff
@@ -0,0 +1,105 @@
...
+  godox:
+    # TODO, OPT, etc. comments are fine to commit. Use FIXME comments for
+    # temporary hacks, and use godox to prevent committing them.
+    keywords: [FIXME]
+  tagliatelle:
+    # Check the struck tag name case.
+    case:
+      rules:
+        json: snake
+        yaml: snake
```
pkwarren, line 34: Added this for you @mfridman.

mfridman: Appreciate it. Although may be overkill as I'm sure this will result in a lot of `//nolint:` because of the upstream JSON blobs we're interfacing with, esp. in fetcher.

mfridman (approved):
This all looks sane, I do like the release portion of it where one can quickly get the current state of the world.

Let's sync up about the .zip files and consider making the docker images public. I did a POC in core to see what uploading a .zip file would look like, and it's not great to be honest. For as long as plugins are docker images, we should lean on the docker ecosystem as much as possible.

PR #244, opened 2022-11-30
add mypy-protobuf and mypy-protobuf-grpc plugins

### LINKED ISSUE #219: Plugin request for Buf Schema Registry: mypy-protobuf
## Mandatory

https://github.com/nipunn1313/mypy-protobuf/tree/main/mypy_protobuf

## Optional

Yes, currently `3.4.0`, from https://github.com/nipunn1313/mypy-protobuf/blob/main/CHANGELOG.md.

I think so, under ["Requirements to run mypy-protobuf"](https://github.com/nipunn1313/mypy-protobuf#requirements-to-run-mypy-protobuf).

Nope - it's commonly used with [`python-protobuf`](https://pypi.org/project/protobuf/), but isn't dependent on it.

N/A, but perhaps a Python registry? 😃

pkwarren:
Create mypy-protobuf and mypy-protobuf-grpc plugins.

Fixes #219.


plugins/community/mypy-protobuf-grpc/v3.4.0/Dockerfile
```diff
@@ -0,0 +1,14 @@
...
+FROM python:3.11.0-alpine3.17 AS build
+WORKDIR /app
+RUN python -mvenv /app
+ADD /requirements.txt requirements.txt
+RUN source ./bin/activate \
+ && pip install --no-cache-dir -r requirements.txt \
+ && pip uninstall --yes pip setuptools \
+ && rm -f requirements.txt bin/activate.fish bin/activate.csh bin/Activate.ps1
+
+FROM python:3.11.0-alpine3.17
```
pkwarren, line 11: The distroless python package is still marked experimental so using alpine for now.

plugins/community/mypy-protobuf-grpc/v3.4.0/requirements.txt
```diff
@@ -0,0 +1,3 @@
+mypy-protobuf==3.4.0
+protobuf==4.21.9
+types-protobuf==4.21.0.1
```
pkwarren, line 3: Created this with `pip freeze` in the container after installing. This should ensure consistent builds.

plugins/community/nipunn1313-mypy-protobuf-grpc/v3.4.0/buf.plugin.yaml
```diff
@@ -1,5 +1,5 @@
 version: v1
-name: buf.build/community/mypy-protobuf-grpc
+name: buf.build/community/nipunn1313-mypy-protobuf-grpc
```
mfridman, line 2: Should this be `nipunn1313-mypy-grpc` ?

pkwarren: They are both sourced from the same repo nipunn1313-mypy-protobuf - the only difference is the entrypoint to the binary: https://github.com/nipunn1313/mypy-protobuf/blob/7692825cfcf7ef82f2428aaa40e78fb0ba449379/setup.cfg#L25

I thought this would work better to make clear that they're both stemming from a common codebase.

mfridman (approved):

PR #248, opened 2022-12-01
collapse previously released plugins

pkwarren:
Update the release notes body to collapse previously released plugins so users can focus on just what changed.


mfridman (approved):
Thank you for adding this.

PR #249, opened 2022-12-02
add a utility to download plugins from GitHub

pkwarren:
Create a download-plugins CLI which can be used to download the latest plugins from GitHub (and verify both the signature of plugin-releases.json and the SHA-256 sums of each zip file).

As part of this work, refactor reusable code from the current release command into the internal/release package.

### DIFF
--- a/cmd/download-plugins/main.go
+++ b/cmd/download-plugins/main.go
@@ -1,0 +1,164 @@
+package main
+
+import (
+	"context"
+	"crypto/sha256"
+	_ "embed"
+	"encoding/hex"
+	"errors"
+	"flag"
+	"fmt"
+	"io"
+	"log"
+	"net/http"
+	"os"
+	"path/filepath"
+	"strings"
+
+	"aead.dev/minisign"
+	"github.com/google/go-github/v48/github"
+
+	"github.com/bufbuild/plugins/internal/release"
+)
+
+func main() {
+	if err := run(); err != nil {
+		log.Fatalf("failed to download release: %v", err)
+	}
+}
+
+func run() error {
+	var (
+		minisignPublicKey string
+		releaseTag        string
+	)
+	flag.StringVar(&minisignPublicKey, "minisign-public-key", "", "path to minisign public key file")
+	flag.StringVar(&releaseTag, "release-tag", "", "release to download (default: latest release)")
+	flag.Parse()
+
+	if len(flag.Args()) != 1 {
+		_, _ = os.Stderr.WriteString("usage: download-plugins <dir>\n")
+		flag.PrintDefaults()
+		os.Exit(2)
+	}
+	downloadDir := flag.Arg(0)
+
+	ctx := context.Background()
+	client := release.NewClient(ctx)
+
+	publicKey, err := loadPublicKey(minisignPublicKey)
+	if err != nil {
+		return err
+	}
+
+	var githubRelease *github.RepositoryRelease
+	if releaseTag == "" {
+		githubRelease, err = client.GetLatestRelease(ctx, release.GithubOwnerBufbuild, release.GithubRepoPlugins)
+		if err != nil {
+			return err
+		}
+	} else {
+		githubRelease, err = client.GetReleaseByTag(ctx, release.GithubOwnerBufbuild, release.GithubRepoPlugins, releaseTag)
+		if err != nil {
+			return err
+		}
+	}
+
+	pluginReleases, err := client.LoadPluginReleases(ctx, githubRelease, publicKey)
+	if err != nil {
+		return err
+	}
+	for _, pluginRelease := range pluginReleases.Releases {
+		exists, err := pluginExistsMatchingDigest(pluginRelease, downloadDir)
+		if err != nil {
+			return err
+		}
+		if exists {
+			log.Printf("already downloaded: %s", filepath.Join(downloadDir, filepath.Base(pluginRelease.URL)))
+		} else if err := downloadReleaseToDir(ctx, client.GitHub.Client(), pluginRelease, downloadDir); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func pluginExistsMatchingDigest(plugin release.PluginRelease, downloadDir string) (bool, error) {
+	digest, err := release.CalculateDigest(filepath.Join(downloadDir, filepath.Base(plugin.URL)))
+	if err != nil {
+		return false, err
+	}
+	return digest == plugin.PluginZipDigest, nil
+}
+
+func downloadReleaseToDir(ctx context.Context, client *http.Client, plugin release.PluginRelease, downloadDir string) error {
+	expectedDigest, err := parseDigest(plugin.PluginZipDigest)
+	if err != nil {
+		return err
+	}
+	f, err := os.CreateTemp(downloadDir, "."+strings.ReplaceAll(plugin.PluginName, "/", "-"))
+	if err != nil {
+		return err
+	}
+	defer func() {
+		if err := f.Close(); err != nil && !errors.Is(err, os.ErrClosed) {
+			log.Printf("failed to close temporary file: %v", err)
+		}
+		if err := os.Remove(f.Name()); err != nil && !errors.Is(err, os.ErrNotExist) {
+			log.Printf("failed to remove temporary file %q: %v", f.Name(), err)
+		}
+	}()
+	log.Printf("downloading: %v", plugin.URL)
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, plugin.URL, nil)
+	if err != nil {
+		return err
+	}
+	resp, err := client.Do(req)
+	if err != nil {
+		return err
+	}
+	defer func() {
+		if err := resp.Body.Close(); err != nil {
+			log.Printf("failed to close response: %v", err)
+		}
+	}()
+	if resp.StatusCode != http.StatusOK {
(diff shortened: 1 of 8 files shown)


cmd/download-plugins/main.go
```diff
@@ -0,0 +1,164 @@
...
+	if err != nil {
+		return err
+	}
+	for _, pluginRelease := range pluginReleases.Releases {
+		exists, err := pluginExistsMatchingDigest(pluginRelease, downloadDir)
+		if err != nil {
+			return err
+		}
+		if exists {
+			log.Printf("already downloaded: %s", filepath.Join(downloadDir, filepath.Base(pluginRelease.URL)))
```
pkwarren, line 77: If plugins are already downloaded to the download dir, we'll skip downloading them again if the SHA-256 digest matches:

```
2022/12/01 21:44:49 [DEBUG] GET https://api.github.com/repos/bufbuild/plugins/releases/latest
2022/12/01 21:44:49 [DEBUG] GET https://api.github.com/repos/bufbuild/plugins/releases/assets/86688755
2022/12/01 21:44:50 [DEBUG] GET https://api.github.com/repos/bufbuild/plugins/releases/assets/86688756
2022/12/01 21:44:50 already downloaded: download/apple-swift-v1.20.2.zip
2022/12/01 21:44:50 already downloaded: download/apple-swift-v1.20.3.zip
2022/12/01 21:44:50 already downloaded: download/bufbuild-connect-go-v1.0.0.zip
...
2022/12/01 21:44:51 already downloaded: download/protocolbuffers-ruby-v21.8.zip
2022/12/01 21:44:51 already downloaded: download/protocolbuffers-ruby-v21.9.zip
2022/12/01 21:44:51 already downloaded: download/protocolbuffers-ruby-v21.10.zip
```

cmd/download-plugins/main.go
```diff
@@ -0,0 +1,164 @@
...
+	}
+	return digest == plugin.PluginZipDigest, nil
+}
+
+func downloadReleaseToDir(ctx context.Context, client *http.Client, plugin release.PluginRelease, downloadDir string) error {
+	expectedDigest, err := parseDigest(plugin.PluginZipDigest)
+	if err != nil {
+		return err
+	}
+	f, err := os.CreateTemp(downloadDir, "."+strings.ReplaceAll(plugin.PluginName, "/", "-"))
```
pkwarren, line 98: Downloads the zip file to a temporary file and then renames it when complete (if digest matches).

cmd/release/main.go
```diff
@@ -26,123 +23,170 @@ import (
...
 
 func main() {
-	var (
-		minisignPrivateKey string
-		dryRun             bool
+	dryRun := flag.Bool("dry-run", false, "perform a dry-run (no GitHub modifications)")
+	githubReleaseOwner := flag.String(
+		"github-release-owner",
+		string(release.GithubOwnerBufbuild),
+		"GitHub release owner (set to personal account to test against a fork)",
```
pkwarren, line 41: Made this a CLI arg so you don't have to modify the code to test against a personal fork.

cmd/release/main.go
```diff
@@ -26,123 +23,170 @@ import (
...
+	)
+	minisignPrivateKey := flag.String("minisign-private-key", "", "path to minisign private key file")
+	minisignPublicKey := flag.String(
+		"minisign-public-key",
+		"",
+		"path to public key used to verify the latest release's plugin-releases.json file (if different than private key)",
```
pkwarren, line 47: This arg is used if the public key used to verify the plugin-releases.json file doesn't match the private key used to sign (this might be if you're using `-dry-run` locally).

cmd/release/main.go
```diff
@@ -26,123 +23,170 @@ import (
...
 
-	var newPlugins []PluginRelease
-	var existingPlugins []PluginRelease
 	now := time.Now().UTC().Truncate(time.Second)
 	releaseName, err := calculateNextRelease(now, latestRelease)
 	if err != nil {
 		return fmt.Errorf("failed to determine next release name: %w", err)
 	}
 
-	if err := plugin.Walk(root, func(plugin *plugin.Plugin) error {
```
pkwarren, line 140: Kept pushing up against the edge of cyclomatic complexity linter so moved the bulk of the run method into `calculateNewReleasePlugins`.

internal/release/github.go
```diff
@@ -0,0 +1,171 @@
...
+	"strings"
+
+	"aead.dev/minisign"
+	"github.com/google/go-github/v48/github"
+	"github.com/hashicorp/go-retryablehttp"
+	"golang.org/x/oauth2"
+)
+
+type GithubRepo string
+type GithubOwner string
```
pkwarren, line 22: Not strictly necessary but I want to avoid passing the wrong thing to the client methods (easy to get things out of order).

cmd/release/main.go
```diff
@@ -183,98 +222,101 @@ func run(root string, minisignPrivateKey string, dryRun bool) error {
...
-func createRelease(ctx context.Context, client *github.Client, releaseName string, plugins []PluginRelease, tmpDir string, publicKey *minisign.PublicKey) error {
 	// Create GitHub release
-	release, _, err := client.Repositories.CreateRelease(ctx, githubReleaseOwner, githubRepo, &github.RepositoryRelease{
+	repositoryRelease, err := client.CreateRelease(ctx, c.githubReleaseOwner, release.GithubRepoPlugins, &github.RepositoryRelease{
 		TagName: github.String(releaseName),
 		Name:    github.String(releaseName),
-		Body:    github.String(createReleaseBody(releaseName, plugins, publicKey)),
+		Body:    github.String(releaseBody),
+		// Start release as a draft until all assets are uploaded
+		Draft: github.Bool(true),
```
pkwarren, line 284: Tweaked the release process to mark it as a draft until all the plugins were uploaded, then change it to an official release at the end. I noticed our notifications of new releases were going out prior to all the assets being uploaded.

internal/release/github.go
```diff
@@ -0,0 +1,171 @@
...
+func (c *Client) LoadPluginReleases(ctx context.Context, release *github.RepositoryRelease, publicKey minisign.PublicKey) (*PluginReleases, error) {
+	releasesJSONBytes, err := c.downloadAsset(ctx, release, PluginReleasesFile)
+	if err != nil {
+		return nil, err
+	}
+	releasesJSONMinisigBytes, err := c.downloadAsset(ctx, release, PluginReleasesSignatureFile)
+	if err != nil && !errors.Is(err, ErrNotFound) {
+		return nil, err
+	}
+	if !publicKey.Equal(minisign.PublicKey{}) && !minisign.Verify(publicKey, releasesJSONBytes, releasesJSONMinisigBytes) {
```
pkwarren, line 136: Now if we're passed a valid `minisign.PublicKey`, we verify it first before parsing it. This means that as long as keys aren't changing, we'll always verify contents before creating a new release based on this.

internal/release/minisign.pub
```diff
@@ -0,0 +1,2 @@
+untrusted comment: minisign public key: 4C9786A7A6FF1D33
```
pkwarren, line 1: This is the minisign public key used to sign the official bufbuild/plugins releases. Added it here so we don't have to try to parse it from release notes or anything like that.

cmd/release/main.go
```diff
@@ -499,12 +537,12 @@ func calculateNextRelease(now time.Time, latestRelease *github.RepositoryRelease
...
 }
 
-func pluginDownloadURL(plugin *plugin.Plugin, releaseName string) (string, error) {
+func (c *command) pluginDownloadURL(plugin *plugin.Plugin, releaseName string) (string, error) {
 	zipName, err := pluginZipName(plugin)
 	if err != nil {
 		return "", err
 	}
-	return fmt.Sprintf("https://github.com/%s/%s/releases/download/%s/%s", githubReleaseOwner, githubRepo, releaseName, zipName), nil
+	return fmt.Sprintf("https://github.com/%s/%s/releases/download/%s/%s", c.githubReleaseOwner, release.GithubRepoPlugins, releaseName, zipName), nil
```
mfridman, line 545: Curious why not use the GH API client?

internal/release/github.go
```diff
@@ -0,0 +1,171 @@
...
+		if resp != nil && resp.StatusCode == http.StatusNotFound {
+			// no latest release
+			return nil, ErrNotFound
+		}
+		return nil, err
+	}
+	return repositoryRelease, nil
+}
+
+func getOrgRepoFromReleaseURL(url string) (string, string, error) {
```
mfridman, line 109: Maybe a personal nit, but it makes it a bit easier to reason about if there's an example comment of the input.

internal/release/github.go
mfridman: Does this mean if we rotate the key, this will break because of a mismatch with previous release?

mfridman (approved):
Looks great!

internal/release/github.go
pkwarren: Yep - if we rotate the key we might need to unblock releases by adding the new key in addition to the old key, and supporting both. Once we've released w/ the new key, we could remove the old key.

cmd/release/main.go
pkwarren: In this case, we're crafting the download URL prior to uploading it via the release API (since it will be part of the `package-releases.json` file). If we wanted to do this after uploading the assets, we'd need to generate the JSON file last and sign it.

internal/release/github.go
pkwarren: Good call - I'll add an example.

PR #254, opened 2022-12-02
Add crates.io API support in fetchclient

### LINKED ISSUE #252: Update fetcher to support Rust-based plugins
The fetcher currently supports the following upstream sources:

- GitHub
- Dart / Flutter
- NPM registry
- Go Proxy directly
- Maven Search

To support Rust-based plugins, such as https://github.com/bufbuild/plugins/issues/96#issuecomment-1324216611, we need a way to fetch those versions.

I'm hoping we can standardize on using crates.io to fetch the semver information.

**Important:** take note of the following:

https://crates.io/policies#crawlers
https://crates.io/data-access

mfridman:
Fix #252

Tested once.

```
2022/12/02 18:16:56 [DEBUG] GET https://crates.io/api/v1/crates/protoc-gen-prost
2022/12/02 18:16:57 v0.2.1
```

### DIFF
--- a/internal/fetchclient/fetchclient.go
+++ b/internal/fetchclient/fetchclient.go
@@ -21,6 +21,7 @@ import (
 )
 
 const (
+	cratesURL = "https://crates.io/api/v1/crates"
 	// docs: https://pub.dev/help/api
 	dartFlutterAPIURL = "https://pub.dev/api/packages"
 	goProxyURL        = "https://proxy.golang.org"
@@ -84,6 +85,8 @@ func (c *Client) fetch(ctx context.Context, config *source.Config) (string, erro
 			return "", err
 		}
 		return results.latestVersion, nil
+	case config.Source.Crates != nil:
+		return c.fetchCrate(ctx, config.Source.Crates.CrateName)
 	}
 	return "", errors.New("failed to match a source")
 }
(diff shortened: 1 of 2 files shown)


internal/fetchclient/fetchclient.go
```diff
@@ -115,6 +118,52 @@ func (c *Client) fetchDartFlutter(ctx context.Context, name string) (string, err
...
+	// See https://github.com/bufbuild/plugins/issues/252 for more information.
+	// We must be careful with this API and respect the crawling policy.
+	request.Header.Set("User-Agent", "bufbuild (github.com/bufbuild/plugins)")
+	response, err := c.httpClient.Do(request)
+	if err != nil {
+		return "", err
+	}
+	defer response.Body.Close()
+
+	var data struct {
```
mfridman, line 140: We might be able to just use `newest_version` or the `max_stable_version` instead of looping through all the versions.

internal/fetchclient/fetchclient.go
```diff
@@ -115,6 +118,52 @@ func (c *Client) fetchDartFlutter(ctx context.Context, name string) (string, err
...
+		} `json:"versions"`
+	}
+	if err := json.NewDecoder(response.Body).Decode(&data); err != nil {
+		return "", err
+	}
+	var versions []string
+	for _, version := range data.Versions {
+		if version.Yanked {
+			// A yanked version a is a published crate's version that has been removed
+			// from the server's index.
```
pkwarren, line 153: If we encounter a crate that was the latest release and then later "yanked", what will be our solution on the BSR? Will we need the ability to disable a plugin so that newer code that doesn't specify a version and remote packages builds with the last stable release?

pkwarren (approved):
Nice!

internal/fetchclient/fetchclient.go
pkwarren: So far we only allow semver (without prerelease IIRC) so looping might be the best option.

internal/fetchclient/fetchclient.go
mfridman: The crate is still available for download, so technically it's still available even upstream.

I think the best we can do is mark it deprecated, and that'll be a manual process.

We could automate this and maybe create a GH issue if we detect this, which gets assigned to a person. But I'm a bit worried about the ongoing maintenance cost of trying to mirror "deprecated" plugin versions.

internal/fetchclient/fetchclient.go
mfridman: Ye, it's good to know these options are available. And there is also `:crate_id/versions` available.

https://github.com/rust-lang/crates.io/blob/8be79196dbcd7fb1388cf016ab128da11c484ed6/src/router.rs#L76-L79

internal/fetchclient/fetchclient.go
mfridman: Tbh I don't have a good answer. Suggestions welcome.

If we can't be consistent, across the ecosystems, then maybe the best thing is to do nothing and leave it to the user?

PR #259, opened 2022-12-05
migrate off deprecated github action

pkwarren:
Migrate from last-successful-commit-action to nx-set-shas. This reflects upstream changes made in tj-actions/changed-files:

https://github.com/tj-actions/changed-files/pull/822

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,17 +17,17 @@ jobs:
     - uses: actions/checkout@v3
       with:
         fetch-depth: 0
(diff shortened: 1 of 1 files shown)


mfridman:
Do we need to update the permissions as well? 

https://github.com/nrwl/nx-set-shas#permissions-in-v2

It says it needs:

```
    permissions:
      contents: 'read'
      actions: 'read'
```

Or will these be included by default?

mfridman (approved):
lgtm. Just one comment wrt permissions.

pkwarren:
I don't think actions are restricted by default on this repo but I can add those in case that is turned on later.

PR #262, opened 2022-12-06
add grpc/gateway plugin

### LINKED ISSUE #152: Plugin request for Buf Schema Registry - grpc-gateway
We should add grpc-gateway and special-case it to generate to a different package, similar to grpc-go. We should have special documentation around how to use grpc-gateway once we do so as well.

grpc-gateway requires:

- The golang base types
- The grpc-go types
- A dependency on googleapis, with the `go_package` values from googleapis used (this is an `except` when used with managed mode)

This means that the grpc-gateway plugin needs to be aware of the package for the golang base types, and the grpc-go types. Normally, this is the same package. However, we've split out grpc-go to be in a different package, so we need to communicate to the grpc-gateway plugin both the golang base type package, and the grpc-gateway package. This will take a change to grpc-gateway, either through a PR or through a patch, similar to grpc-go.

grpc-gateway recently added a `standalone=true` option, which when used, will generate grpc-gateway code to a different package than the golang base types and the grpc-go types. This is 90% of the battle - the general code to generate grpc-gateway types to a different package is already there. However, it still assumes the golang base types and grpc-go types are all in the same package. We would need to make it so that we can specify a seperate grpc-go type package.

If we do this, grpc-gateway will work with the go module proxy out of the box.

pkwarren:
Add a grpc/gateway plugin.

Fixes #152.


plugins/grpc/gateway/v2.14.0/remote-packages.patch
```diff
@@ -0,0 +1,233 @@
+diff --git a/internal/descriptor/registry.go b/internal/descriptor/registry.go
```
mfridman, line 1: Once we confirm this is working as intended and passes our internal tests, are we planning to merge this as a patch .. or do we wait to get this merged upstream with @johanbrandhorst help.

johanbrandhorst: I'm looking at the patch upstream but I admit it is a very specific use case, and if patching like this is an acceptable solution to you (at least in the short term) then I would prefer this patch stays here. I'm happy to consider upstreaming it if this turns into a big maintenance burden for you down the road. How does that sound?

plugins/grpc-ecosystem/gateway/v2.15.0/.dockerignore
```diff
@@ -0,0 +1,3 @@
+*
+!Dockerfile
+!remote-packages.patch
```
mfridman, line 3: Let's remove this.

mfridman (approved):

plugins/grpc-ecosystem/gateway/v2.15.0/buf.plugin.yaml
```diff
@@ -0,0 +1,12 @@
+version: v1
+name: buf.build/grpc-ecosystem/gateway
```
mfridman, line 2: Okay, I had second thoughts on this, but because buf.build/grpc-ecosystem/grpc-gateway already exists as a module, we'll need to name this something else.

PR #265, opened 2022-12-08
fix download command for new plugins

pkwarren:
Check if the downloaded plugin exists before checking its digest.

### DIFF
--- a/cmd/download-plugins/main.go
+++ b/cmd/download-plugins/main.go
@@ -84,8 +84,7 @@ func run() error {
 
 func pluginExistsMatchingDigest(plugin release.PluginRelease, downloadDir string) (bool, error) {
 	filename := filepath.Join(downloadDir, filepath.Base(plugin.URL))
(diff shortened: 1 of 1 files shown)


cmd/download-plugins/main.go
```diff
@@ -83,21 +83,28 @@ func run() error {
...
+	filename := filepath.Join(downloadDir, filepath.Base(plugin.URL))
+	if _, err := os.Stat(filename); err != nil {
+		if os.IsNotExist(err) {
+			return false, nil
+		}
 		return false, err
 	}
+	digest, err := release.CalculateDigest(filename)
+	if err != nil {
+		return false, fmt.Errorf("failed to calculate digest for plugin %s: %w", plugin.URL, err)
```
mfridman, line 95: Should the error reference the `plugin.URL` or `filename`, which is easier to debug?

mfridman (approved):

PR #267, opened 2022-12-09
small improvements to download-plugins

pkwarren:
Remove DEBUG logging of every HTTP request. Don't log when the plugin exists on disk with the right checksum. Set plugin timestamps to reflect when they were released (this may make it easier to determine which plugins to install or update).

### DIFF
--- a/cmd/download-plugins/main.go
+++ b/cmd/download-plugins/main.go
@@ -32,7 +32,7 @@ func run() error {
 		minisignPublicKey string
 		releaseTag        string
 	)
-	flag.StringVar(&minisignPublicKey, "minisign-public-key", "", "path to minisign public key file")
+	flag.StringVar(&minisignPublicKey, "minisign-public-key", "", "path to minisign public key file (default: bufbuild/plugins public key)")
 	flag.StringVar(&releaseTag, "release-tag", "", "release to download (default: latest release)")
 	flag.Parse()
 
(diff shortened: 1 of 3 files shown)


cmd/download-plugins/main.go
```diff
@@ -32,7 +32,7 @@ func run() error {
 		minisignPublicKey string
 		releaseTag        string
 	)
-	flag.StringVar(&minisignPublicKey, "minisign-public-key", "", "path to minisign public key file")
+	flag.StringVar(&minisignPublicKey, "minisign-public-key", "", "path to minisign public key file (default: bufbuild/plugins public key)")
```
pkwarren, line 35: Useful to mention that there is a default public key used to verify releases.

cmd/download-plugins/main.go
```diff
@@ -73,15 +73,23 @@ func run() error {
 		if err != nil {
 			return err
 		}
-		if exists {
-			log.Printf("already downloaded: %s", filepath.Join(downloadDir, filepath.Base(pluginRelease.URL)))
```
pkwarren, line 77: This log statement is too chatty - better to just display what was downloaded.

cmd/download-plugins/main.go
```diff
@@ -73,15 +73,23 @@ func run() error {
...
-		} else if err := downloadReleaseToDir(ctx, client.GitHub.Client(), pluginRelease, downloadDir); err != nil {
+		if !exists {
+			if err := downloadReleaseToDir(ctx, client.GitHub.Client(), pluginRelease, downloadDir); err != nil {
+				return err
+			}
+		}
+		if err := setPluginZipTimestamp(pluginRelease, downloadDir); err != nil {
```
pkwarren, line 81: This felt like a nice change to make file's modified time on disk match when they were released.

internal/fetchclient/fetchclient.go
```diff
@@ -47,7 +47,9 @@ func New(ctx context.Context) *Client {
 		)
 		client = oauth2.NewClient(ctx, ts)
 	} else {
-		client = retryablehttp.NewClient().StandardClient()
+		retryableClient := retryablehttp.NewClient()
+		retryableClient.Logger = nil
```
pkwarren, line 51: Works around: https://github.com/hashicorp/go-retryablehttp/issues/31

mfridman (approved):

PR #272, opened 2022-12-12
check http status code of responses

pkwarren:
Update the fetch client to check the status code of responses before trying to parse them. Log more details about errors for the Maven fetch implementation as we're seeing regular issues there and it would help to diagnose further (perhaps rate limiting?).


mfridman (approved):
Is it primarily the maven search that's consistently flaky? 

In https://github.com/bufbuild/plugins/pull/191 we started iterating over the entire set, but maybe the last 20 is good enough for our purposes? Either way, let's see what the error is next time it happens.

PR #277, opened 2022-12-15
switch fetcher to use maven-metadata.xml

pkwarren:
Instead of dealing w/ the rest API (and pagination), simply fetch the maven-metadata.xml file and determine versions from there.

### DIFF
--- a/internal/fetchclient/fetchclient.go
+++ b/internal/fetchclient/fetchclient.go
@@ -3,15 +3,13 @@ package fetchclient
 import (
 	"context"
 	"encoding/json"
+	"encoding/xml"
 	"errors"
 	"fmt"
-	"io"
 	"log"
-	"math"
 	"net/http"
 	"net/url"
 	"os"
-	"strconv"
 	"strings"
 
 	"github.com/google/go-github/v48/github"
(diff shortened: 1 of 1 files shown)


internal/fetchclient/fetchclient.go
```diff
@@ -321,17 +247,45 @@ func (c *Client) fetchMavenPage(ctx context.Context, targetURL *url.URL, group s
...
+			version = "v" + version
+		}
+		if !semver.IsValid(version) || semver.Prerelease(version) != "" {
+			continue
+		}
+		version = semver.Canonical(version)
+		if results.latestVersion == "" || semver.Compare(results.latestVersion, version) < 0 {
+			results.latestVersion = version
+		}
+		release := semver.MajorMinor(version)
```
mfridman, line 280: Can you remind me why we're doing this extra bit of logic for maven in addition to the semver.Compare above?

pkwarren: The hope was that for older versions of plugins (that have new patch releases available for security issues) that we'd also discover those as new plugin versions in addition to the latest version changing.

mfridman: At the end we only surface a single version, so if we have [v1.2.3, v1.2.4, v1.3.0] then v1.3.0 will get returned (even if v1.2.4 was also included). The initial sort should ensure the latest semver gets returned?

I'm also not convinced we need this extra logic, because we only care about the latest semver. I.e., if you're on v1.2.3 and v1.3.0 is available, why do you need v1.2.4? If this is an issue for end-users we could always backfill a version, but I suspect for 99% of users the latest should suffice?

mfridman: I think the thing that confuses me is that we never use `latestPatchVersions`, so every time I stumble on the maven fetch this logic trips me up 😅 . Maybe you can add a comment as to why this is different to all the other fetchers?

mfridman (approved):

PR #278, opened 2022-12-16
add missing plugin dependencies

pkwarren:
Even though these plugins don't yet support remote packages, we should add dependencies so that the 'Include Dependencies' works as expected on the plugin details pages. All of these plugins require base types generated from another plugin.

### DIFF
--- a/plugins/bufbuild/validate-cpp/v0.9.0/buf.plugin.yaml
+++ b/plugins/bufbuild/validate-cpp/v0.9.0/buf.plugin.yaml
@@ -3,6 +3,8 @@ name: buf.build/bufbuild/validate-cpp
 plugin_version: v0.9.0
 source_url: https://github.com/bufbuild/protoc-gen-validate
 description: Generates C++ code to validate Protobuf messages using protoc-gen-validate constraints.
+deps:
+  - plugin: buf.build/protocolbuffers/cpp:v21.12
 output_languages:
   - cpp
 spdx_license_id: Apache-2.0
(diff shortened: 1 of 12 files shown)


plugins/bufbuild/validate-cpp/v0.9.0/buf.plugin.yaml
```diff
@@ -3,6 +3,8 @@ name: buf.build/bufbuild/validate-cpp
 plugin_version: v0.9.0
 source_url: https://github.com/bufbuild/protoc-gen-validate
 description: Generates C++ code to validate Protobuf messages using protoc-gen-validate constraints.
+deps:
+  - plugin: buf.build/protocolbuffers/cpp:v21.12
```
pkwarren, line 7: I started by pinning the latest available version. We can update when new versions of these plugins are updated (but I don't think it is worth updating each time a dependency updates).

plugins/grpc/cpp/v1.51.1/buf.plugin.yaml
```diff
@@ -3,6 +3,8 @@ name: buf.build/grpc/cpp
 plugin_version: v1.51.1
 source_url: https://github.com/grpc/grpc
 description: Generates C++ client and server stubs for the gRPC framework.
+deps:
+  - plugin: buf.build/protocolbuffers/cpp:v21.12
```
pkwarren, line 7: I didn't backfill these changes back through all the older versions of these plugins - just the latest version.

mfridman (approved):

PR #279, opened 2022-12-18
add infobloxopen-gorm v1.1.2

tboerger:
Generate GORM models based on proto definitions :)


plugins/community/infobloxopen-gorm/v1.1.2/Dockerfile
```diff
@@ -0,0 +1,11 @@
+# syntax=docker/dockerfile:1.4
+FROM golang:1.19.2-bullseye AS build
```
pkwarren, line 2:
```suggestion
FROM golang:1.19.4-bullseye AS build
```

When creating new plugins, we update to the latest base images.

plugins/community/infobloxopen-gorm/v1.1.2/buf.plugin.yaml
```diff
@@ -0,0 +1,7 @@
+version: v1
+name: buf.build/community/infobloxopen-gorm
+plugin_version: v1.1.2
+source_url: https://github.com/infobloxopen/protoc-gen-gorm
+description: Generate GORM models and APIs for simple object persistence tasks.
+spdx_license_id: Apache-2.0
+license_url: https://github.com/infobloxopen/protoc-gen-gorm/blob/v1.1.2/LICENSE
```
pkwarren, line 7: Can you add an `output_languages` section too?

```yaml
output_languages:
  - go
```

tests/testdata/buf.build/community/infobloxopen-gorm/v1.1.2/eliza/plugin.sum
```diff
@@ -0,0 +1 @@
+h1:47DEQpj8HBSa+/TImW+5JCeuQeRkm5NMpJWZG3hSuFU=
```
pkwarren, line 1: I haven't used this plugin before and our test inputs don't have the required options (`gorm.*`) specified, so currently there is no code generated with our tests. We might need to manually verify the plugin against a more complete example to ensure it will do the right thing when added to the BSR.

plugins/community/infobloxopen-gorm/v1.1.2/Dockerfile
tboerger: Updated to new version

plugins/community/infobloxopen-gorm/v1.1.2/buf.plugin.yaml
tboerger: Added this key

tests/testdata/buf.build/community/infobloxopen-gorm/v1.1.2/eliza/plugin.sum
tboerger: Is there something I can provide?

plugins/community/infobloxopen-gorm/v1.1.2/buf.plugin.yaml
```diff
@@ -0,0 +1,9 @@
...
+output_languages:
+  - go
```
pkwarren, line 9: Can you also add this:

```yaml
deps:
  - plugin: buf.build/protocolbuffers/go:v1.28.1
```

It looks like the generated code depends on the Go generated base types, so this will help provide a more complete buf.gen.yaml example in the UI (by listing dependencies).

tests/testdata/buf.build/community/infobloxopen-gorm/v1.1.2/eliza/plugin.sum
bufdev: You'd need to modify the plugin so that the base case works when no options are specified - this should not result in an error (perhaps it results in no code output)

bufdev:
Optimally, the plugin would be modified so that the generated code can live in a different package than the base types, i.e the `.pb.gorm.go` files can live in a different Golang package than the `pb.go` files. Without this, this plugin will not work with remote packages.

bufdev:
If you are using https://godocs.io/google.golang.org/protobuf/compiler/protogen to author your plugin, you can easily figure out the go package for your base types, and then use them for your helper plugin when generating code. https://github.com/bufbuild/connect-go/blob/main/cmd/protoc-gen-connect-go/main.go is an example plugin that does not generate to the same package as the base types.

PR #283, opened 2022-12-19
conditional login to dockerhub

pkwarren:
PRs from GitHub forks don't have access to GitHub actions secrets.

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -16,6 +16,7 @@ jobs:
   pr:
     env:
       BUILDKIT_PROGRESS: plain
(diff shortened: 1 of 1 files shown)


.github/workflows/pr.yml
```diff
@@ -56,7 +57,7 @@ jobs:
         driver: docker
         driver-opts: network=host
     - name: Login to Docker Hub
-      if: github.repository == 'bufbuild/plugins'
```
pkwarren, line 59: I had thought this would be sufficient but this didn't work for PRs from forks.

mfridman (approved):

PR #289, opened 2023-01-03
Update prost source_url to GitHub

mfridman:
This PR updates the source_url to GitHub for the prost plugin to be consistent with other plugins.

### DIFF
--- a/plugins/community/neoeinstein-prost/v0.2.1/buf.plugin.yaml
+++ b/plugins/community/neoeinstein-prost/v0.2.1/buf.plugin.yaml
@@ -1,7 +1,7 @@
 version: v1
 name: buf.build/community/neoeinstein-prost
 plugin_version: v0.2.1
-source_url: https://crates.io/crates/protoc-gen-prost
+source_url: https://github.com/neoeinstein/protoc-gen-prost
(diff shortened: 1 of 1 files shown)


plugins/community/neoeinstein-prost/v0.2.1/buf.plugin.yaml
```diff
@@ -1,7 +1,7 @@
 version: v1
 name: buf.build/community/neoeinstein-prost
 plugin_version: v0.2.1
-source_url: https://crates.io/crates/protoc-gen-prost
+source_url: https://github.com/neoeinstein/protoc-gen-prost
```
mfridman, line 4: There is an argument to be made that the correct source is the plugin installation, and not the GitHub repository.

pkwarren (approved):

plugins/community/neoeinstein-prost/v0.2.1/buf.plugin.yaml
mfridman: I wonder if in addition to `source_url` we can have a `homepage` that links to the npm registry, go proxy docs, crates.io, etc.

PR #290, opened 2023-01-03
Adds protobuf-ts v2.8.2

### LINKED ISSUE #256: Plugin request for Buf Schema Registry - timostamm/protobuf-ts
## Mandatory

https://github.com/timostamm/protobuf-ts

## Optional

v2.8.2 on BSR registry

[@protobuf-ts/runtime](https://www.npmjs.com/package/@protobuf-ts/runtime)

no

yes

mfridman:
Fixes #256


plugins/community/timostamm-protobuf-ts/v2.8.2/buf.plugin.yaml
```diff
@@ -0,0 +1,17 @@
...
+description: Generates Protobuf and RPC for TypeScript.
+output_languages:
+  - typescript
+registry:
+  npm:
+    import_style: module
+    deps:
+      - package: '@protobuf-ts/runtime'
+        version: ^2.8.2
+      - package: '@protobuf-ts/runtime-rpc'
```
mfridman, line 14: I believe `@protobuf-ts/runtime-rpc` is also needed.. although I don't see that in the issue:

https://github.com/bufbuild/plugins/issues/256#issue-1473654420

cc @timostamm

plugins/community/timostamm-protobuf-ts/v2.8.2/buf.plugin.yaml
```diff
@@ -0,0 +1,17 @@
+version: v1
+name: buf.build/community/timostamm-protobuf-ts
+plugin_version: v2.8.2
+source_url: https://github.com/timostamm/protobuf-ts
+description: Generates Protobuf and RPC for TypeScript.
+output_languages:
+  - typescript
+registry:
+  npm:
+    import_style: module
```
mfridman, line 10: Iirc this should just work because it's the plugin default, there is a `output_legacy_commonjs` .. but we do not need to concern ourselves with that.

plugins/community/timostamm-protobuf-ts/v2.8.2/buf.plugin.yaml
```diff
@@ -0,0 +1,17 @@
+version: v1
+name: buf.build/community/timostamm-protobuf-ts
```
mfridman, line 2: The import is a tad long: `@buf/acme_paymentapis.community_timostamm-protobuf-ts`

Not suggesting we change it, but hopefully most users will copy/paste this.

plugins/community/timostamm-protobuf-ts/v2.8.2/buf.plugin.yaml
```diff
@@ -0,0 +1,17 @@
...
+plugin_version: v2.8.2
+source_url: https://github.com/timostamm/protobuf-ts
+description: Generates Protobuf and RPC for TypeScript.
+output_languages:
+  - typescript
```
pkwarren, line 7: It looks like it also has support for emitting JavaScript via options so we should add that here too.

plugins/community/timostamm-protobuf-ts/v2.8.2/Dockerfile
```diff
@@ -0,0 +1,11 @@
+# syntax=docker/dockerfile:1.4
+FROM node:19.3.0-alpine3.17 AS build
```
pkwarren, line 2: Should we stick to a node LTS version instead of a dev version?

plugins/community/timostamm-protobuf-ts/v2.8.2/buf.plugin.yaml
```diff
@@ -0,0 +1,17 @@
...
+registry:
```
pkwarren, line 8: Did you test remote packages support?

plugins/community/timostamm-protobuf-ts/v2.8.2/Dockerfile
mfridman: Ye we probably should, let me change that to be consistent.

plugins/community/timostamm-protobuf-ts/v2.8.2/buf.plugin.yaml
timostamm: This is a bit complicated, because the required runtime libraries actually change, depending on the plugin options being used.

For example, with the plugin option `client_grpc1` or `server_grpc1`, the generated code requires `@grpc/grpc-js`.
With the plugin option `server_generic` or `client_generic` (set by default), `@protobuf-ts/runtime-rpc` is required.

There are also custom options to change behavior per service, for example:

```proto
syntax = "proto3";
import "protobuf-ts.proto";
import "google/protobuf/wrappers.proto";
service S {
    option (ts.client) = GRPC1_CLIENT;
    rpc Hello (google.protobuf.StringValue) returns (google.protobuf.Int32Value);
}
```

So there is no easy answer for this. I think the most frequently used option is `client_generic`, and it's also the default, so there is an argument for adding `@protobuf-ts/runtime-rpc` to the runtime libraries. 

WDYT?

plugins/community/timostamm-protobuf-ts/v2.8.2/buf.plugin.yaml
timostamm: For best remote packages support, the option `output_javascript` should be set. This will output JavaScript files along with TypeScript declaration files. This is very heavy on resources though.

plugins/community/timostamm-protobuf-ts/v2.8.2/buf.plugin.yaml
mfridman: Afaics the existing template had the following options:

```
output_javascript,optimize_code_size,long_type_string,add_pb_suffix,ts_nocheck,eslint_disable
```

What would you recommend as the default options when using this as part of remote packages? Tbh I don't have a strong opinion here so looking for guidance to satisfy the most common case.

mfridman: To keep parity with the old template, I blanket copied the options and runtime libraries in https://github.com/bufbuild/plugins/pull/290/commits/ef529f78b890ad22e2f8b6e2f417a995dc0c96fb. Seems like this is a good start.

Does this look sane to you @timostamm ?

timostamm (approved):
> Does this look sane to you @timostamm ?

Yes, thank you for this!

mfridman:
> > Does this look sane to you @timostamm ?
> 
> Yes, thank you for this!

It's the least I can do, besides buying you a coffee 😄

pkwarren (approved):

Issue #96, opened 2022-09-20
Plugin request for Buf Schema Registry - prost plugins


mfridman:
Do we want any of these?

https://buf.build/prost/plugins

- prost
- serde
- tonic
- crate

neoeinstein:
As the author of these plugins, let me know if there's anything I can do to help, or that would contribute to the goals here.

mfridman:
Hey @neoeinstein, thanks for dropping a comment. 

Any chance you could link us to the source code for these plugin(s)?

Also, do these plugins operate like standard Protobuf plugins that take a `CodeGeneratorRequest` on stdin and return a `CodeGeneratorResponse` on stdout? Apologies, it'd been a while since I've looked at these..

Part of open-sourcing this repository and packaging these plugins is for transparency and exposing more information to end-users about the plugins, their documentation, source, etc.

neoeinstein:
Indeed, they do. The development for these plugins is at [neoeinstein/protoc-gen-prost](https://github.com/neoeinstein/protoc-gen-prost)

I haven't set up a _full_ docker build for these just yet, but it shouldn't be to difficult for me to make that happen. Generally, I have been using `cross` to cross-compile the source for musl and then adding the compiled binaries into a `scratch` docker image, but I could look at setting up a Dockerfile that compiles each of the plugins completely inside a Dockerfile.

neoeinstein:
Not the fastest build in the world, but I have a `Dockerfile` prepared that can be run serially to produce all four of the plugins. The first run will take the lion's share of time, but each plugin after that should be relatively fast (since the workspace can share the same base docker images). On my host, it was 7m, 49s, 45s, and 1m44s respectively from nothing. See the linked PR for the updated `Dockerfile` and `publish_buf.sh` scripts.

mfridman:
Awesome, I'll try to take a look into this as soon as I get some cycles.

mfridman:
Hey @neoeinstein, what are your thoughts wrt the versioning question? One requirement we currently have for managed plugins is valid semver.

I took a brief look and it seems like you're publishing to crates.io? So technically we can fetch and package the plugins from there, right?

https://crates.io/search?q=protoc-gen-prost

neoeinstein:
[...]

PR #300, opened 2023-01-09
Adds protoc-gen-tonic and protoc-gen-prost-serde

mfridman:
Closes https://github.com/bufbuild/plugins/issues/96 by adding:

- [protoc-gen-tonic](https://crates.io/crates/protoc-gen-tonic)
- [protoc-gen-prost-serde](https://crates.io/crates/protoc-gen-prost-serde)

- [ ] Fix tests for plugins that rely on insertion points. Picking up the discussion in:
          - https://github.com/bufbuild/plugins/pull/255#issuecomment-1336485652
          - https://github.com/bufbuild/plugins/pull/255#issuecomment-1343014005

### DIFF
--- a/plugins/community/neoeinstein-prost-serde/source.yaml
+++ b/plugins/community/neoeinstein-prost-serde/source.yaml
@@ -1,0 +1,3 @@
+source:
+  crates:
+    crate_name: protoc-gen-prost-serde
--- a/plugins/community/neoeinstein-prost-serde/v0.2.1/.dockerignore
+++ b/plugins/community/neoeinstein-prost-serde/v0.2.1/.dockerignore
@@ -1,0 +1,2 @@
+*
+!Dockerfile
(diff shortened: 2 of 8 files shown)


mfridman:
I almost wonder if the simplest, and least error-prone solution here is to include:

```yaml
deps:
  - plugin: buf.build/community/neoeinstein-prost:v0.2.1
```

And then change the logic in the test to:

1. Build the plugin and the deps together
2. Modify the `buf.gen.yaml` to include those deps:

https://github.com/bufbuild/plugins/blob/201793406013425c528790c9da8dfcaba86b8c1c/tests/plugins_test.go#L29

pkwarren:
> This would apply to all plugins, and will increase overall time to test a single plugin change. But this seems like a good property anyways, where we test the plugin with the deps. Bonus that it solves the plugin insertion problem.

I like this - seems like the most elegant solution and better reflects the behavior of plugins when run through remote packages.

neoeinstein:
Note, protoc-gen-prost, -prost-serde, and -tonic have now published v0.2.2. Should I still look to publish a plugin to the buf plugin registry as I had previously done?

pkwarren:
> Note, protoc-gen-prost, -prost-serde, and -tonic have now published v0.2.2. Should I still look to publish a plugin to the buf plugin registry as I had previously done?

Sorry for the delay @neoeinstein - we had to work around an issue in our tests with plugins that rely on insertion points. v0.2.2 of all of those plugins are now supported and should be rolling out to the public BSR shortly. You no longer need to worry about pushing them to the BSR - we'll maintain them going forward in this repo.

pkwarren:
Fixed with #401.

Issue #303, opened 2023-01-14
Plugin request for Buf Schema Registry - ZIO GRPC


11Dimensions:
**Where is the source code for the plugin?**

https://github.com/scalapb/zio-grpc

**Does the plugin have a valid semver version?**
Yes

What is the latest version, and where did you get this value from?
v0.5.3 from Github/Maven

**Does this plugin have runtime dependencies?**

This has a runtime dependency of Java

**Does the plugin have a dependency on another plugin?**

No

**Do you think this plugin will be compatible with the BSR go module proxy or npm registry?**

I believe it should be compatible with BSR

PR #307, opened 2023-01-14
ZIO GRPC Scala Plugin

11Dimensions:
Resolves: #303


11Dimensions:
Hi @pkwarren, I think the lack of use is due to a lack of guides of use in the scala community, I was planning on contributing back these guides once these got merged. There were a number of threads where people were curious as to how to use buf with scala that kind of petered out as roadblocks appeared.

I'm not sure how frequently buf would be used in the scala community if the plugins for the most popular effects frameworks are not included. AFAIK no one uses the raw netty implementation (outside of prototyping) baked into the language plugin.

Lastly the maintenance burden shouldn't be more than the core language plugin as all of these libraries follow the same build process provided by the author of scalapb.

pkwarren:
Hey @11Dimensions - that's good feedback. I'll bring this to the team for additional discussion. One concern we had was supporting 3 separate gRPC plugins for Scala, where there is a clear winner for other platforms. I think it might be more reasonable to support 1 gRPC plugin and the validate plugin for example for parity with other language offerings.

I checked the original issues and there weren't additional votes for the plugins from other users and checked our usage stats for custom plugins on the older plugin system and it didn't look like there was sufficient demand to add these to the BSR at this time.

I appreciate the time you've contributed to this project and your feedback. We'll stay in touch with any updates.

11Dimensions:
I'd say that the validate plugin, and the zio plugin would probably be fair seeing as they are all maintained by the same person (who is the one who also maintains the language plugin). So if there are differences that arise due to build changes they would likely affect all projects under their care.

Cats (Typelevel) is also a popular ecosystem - but maybe that can follow on as demand (or the lack there of) determines if adding that plugin is useful. Just for my knowledge, would the calculous change if maintainers of the original project helped in some form or fashion in maintaining the remote plugin?

pkwarren:
We've revisited this request some and decided to reconsider. I've opened #436 with your original PR and a few additional commits on top to pass the CI build.

PR #308, opened 2023-01-14
Create scalapb-validate plugin

### LINKED ISSUE #304: Plugin request for Buf Schema Registry - Scalapb Validate
## Mandatory

https://github.com/scalapb/scalapb-validate

## Optional

v0.3.3 - Github/Maven

This has a runtime dependency on Java

No

It should be compatible with BSR

11Dimensions:
Resolves #304 

Make test command doesn't result in a gen file and the test fails. I believe something is going wrong internally but I'm not sure how to introspect.

Running this built container in the alpha release results in messages like:
`Failure: internal error, contact support, reference code: e311f1`

### DIFF
--- a/plugins/community/scalapb-validate/source.yaml
+++ b/plugins/community/scalapb-validate/source.yaml
@@ -1,0 +1,4 @@
+source:
+  github:
+    owner: scalapb
+    repository: scalapb-validate
--- a/plugins/community/scalapb-validate/v0.3.3/.dockerignore
+++ b/plugins/community/scalapb-validate/v0.3.3/.dockerignore
@@ -1,0 +1,2 @@
+*
+!Dockerfile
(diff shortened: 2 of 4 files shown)


pkwarren:

11Dimensions:
This is a common extension for generating code for the validation plugin, most projects I interact with use this plugin as it converts the types to those described by the plugin.


Additionally this represents a case in which buf plugins hard fail and error out with a opaque error code. Use may be a function of this as I've yet to get this plugin to work in the buf ecosystem (plugin or otherwise)

pkwarren:
Here's the error we get from the plugin execution - I'll update the tests to make it more visible so we can more easily catch this sort of thing in the future:

```
tests/testdata/buf.build/community/scalapb-validate/v0.3.3/eliza$ buf generate ../../../../../images/eliza.bin.gz 
Failure: plugin scalapb-validate: com/buf/connect/demo/eliza/v1/eliza/SayRequest.scala: does not exist
```

11Dimensions:
It's been a while since I've tried it, but I believe I did run them together locally. I'm currently on vacation, but I have an old message that occurred when I published and attempted to use the plugin. Debug didn't give back anything interesting either.

```
 Failure: internal error, contact support, reference code: 090da3
```

PR #312, opened 2023-01-15
Akka GRPC

### LINKED ISSUE #306: Plugin request for Buf Schema Registry - Akka GRPC

https://github.com/akka/akka-grpc/

Of note, not sure how problematic the license is, this is a recent change:
https://github.com/akka/akka-grpc/blob/main/LICENSE

If this is a problem there are earlier versions that need to be imported anyway (people are not updating due to the license change)

Yes, v2.2.1 from Github/Maven

Yes, this depends on Java

No

It should be compatible with BSR

11Dimensions:
Resolves #306 

Added both latest Akka, and latest Apache 2.0 licensed Akka, as there are community members who do not wish to opt into the new license

### DIFF
--- a/plugins/community/lightbend-akka-grpc/source.yaml
+++ b/plugins/community/lightbend-akka-grpc/source.yaml
@@ -1,0 +1,4 @@
+source:
+  github:
+    owner: akka
+    repository: akka-grpc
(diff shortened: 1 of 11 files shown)


pkwarren:

onmomo:
We would be very interested to leverage this plugin as we want to migrate all our service proto generation to a central repository powered by buf. Not having a such a plugin is a blocker for us.

PR #313, opened 2023-01-15
FS2 GRPC Plugin

### LINKED ISSUE #305: Plugin request for Buf Schema Registry - FS2 GRPC
https://github.com/typelevel/fs2-grpc

Yes, v2.5.6 from Github/Maven

Depends on Java

No

Should be compatible with BSR

11Dimensions:
Resolves #305


pkwarren:

PR #319, opened 2023-01-18
fix fetch versions for grpc-gateway

pkwarren:
Use the same configuration as protoc-gen-openapiv2 (which comes from the same repository).

### DIFF
--- a/plugins/grpc-ecosystem/gateway/source.yaml
+++ b/plugins/grpc-ecosystem/gateway/source.yaml
@@ -1,3 +1,4 @@
 source:
-  goproxy:
-    name: github.com/grpc-ecosystem/grpc-gateway/v2/protoc-gen-grpc-gateway
+  github:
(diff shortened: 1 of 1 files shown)


plugins/grpc-ecosystem/gateway/source.yaml
```diff
@@ -1,3 +1,4 @@
 source:
-  goproxy:
-    name: github.com/grpc-ecosystem/grpc-gateway/v2/protoc-gen-grpc-gateway
+  github:
+    owner: grpc-ecosystem
+    repository: grpc-gateway
```
pkwarren, line 4: This should fix https://github.com/bufbuild/plugins/actions/runs/3944373335/jobs/6750199838

stefanvanburen (approved):

mfridman (approved):

PR #329, opened 2023-01-26
Add plugin dependencies to releases file

pkwarren:
In order to optimize lookups, we should add dependency information to the plugin-releases.json file. This will allow us to avoid scanning every .zip file to parse the buf.plugin.yaml.

### DIFF
--- a/cmd/release/main.go
+++ b/cmd/release/main.go
@@ -220,10 +220,12 @@ func (c *command) calculateNewReleasePlugins(currentRelease *release.PluginRelea
 				URL:              downloadURL,
 				LastUpdated:      now,
 				Status:           status,
+				Dependencies:     pluginDependencies(plugin),
 			})
 		} else {
 			log.Printf("plugin %s:%s unchanged", pluginRelease.PluginName, pluginRelease.PluginVersion)
 			pluginRelease.Status = release.StatusExisting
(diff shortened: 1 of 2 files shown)


cmd/release/main.go
```diff
@@ -242,6 +244,21 @@ func (c *command) calculateNewReleasePlugins(currentRelease *release.PluginRelea
...
 }
 
+func pluginDependencies(plugin *plugin.Plugin) []string {
+	if len(plugin.Deps) == 0 {
+		return nil
+	}
+	deps := make([]string, len(plugin.Deps))
+	for i, dep := range plugin.Deps {
+		if dep.Revision != 0 {
+			log.Fatalf("unsupported plugin dependency revision: %v", dep.Revision)
```
pkwarren, line 254: I wish we hadn't modeled revisions in plugin dependencies - they don't make much sense. This will just ensure we don't introduce them in the future.

mfridman: Maybe. If there are 3 revisions of the same plugin version, and I want to ensure I'm using the same exact docker image, how can this be done without an identifier separate from the plugin version?

mfridman: Using latest probably would have been good enough 🤷 , so I definitely see your point.

mfridman (approved):

PR #330, opened 2023-01-26
Add connect-query v0.1.0 plugin

### LINKED ISSUE #328: connect-query
## Mandatory

https://github.com/bufbuild/connect-query/tree/main/packages/protoc-gen-connect-query

## Optional

`package.json`: `"version": "0.1.0"`

Runtime libraries [@bufbuild/connect-query](https://www.npmjs.com/package/@bufbuild/connect-query), and [@bufbuild/protobuf](https://www.npmjs.com/package/@bufbuild/protobuf).

Works in tandem with [@bufbuild/protoc-gen-es](https://www.npmjs.com/package/@bufbuild/protoc-gen-es)

Yes 🎉 

Just refactored our grpc calls to use this! So far the only hiccups are that I needed to make a unique namespaced export because we are generating these files for an private internal library. 

To work around this not being available in the registry for a Docker CI build, is used a bind mount in a Multistage docker build file:
```Dockerfile

FROM bufbuild/buf:1.12.0 as buf

FROM node:18.13.0-alpine
WORKDIR /app
COPY --link . .
RUN npm version "$(cat .version.txt)" && \
  npm i

RUN --mount=type=bind,from=buf,source=/usr/local/bin/buf,target=/usr/local/bin/buf npm run generate # "generate": "buf generate"
```

pkwarren:
Create a plugin for bufbuild/connect-query.

Fixes #328.

### DIFF
--- a/plugins/bufbuild/connect-query/v0.1.0/buf.plugin.yaml
+++ b/plugins/bufbuild/connect-query/v0.1.0/buf.plugin.yaml
@@ -4,7 +4,7 @@ plugin_version: v0.1.0
 source_url: https://github.com/bufbuild/connect-query
 description: Generates client stubs for calling Connect services with TanStack Query.
 deps:
-  - plugin: buf.build/bufbuild/connect-web:v0.6.0
+  - plugin: buf.build/bufbuild/es:v1.0.0
 output_languages:
   - javascript
   - typescript
--- a/plugins/bufbuild/connect-query/v0.1.0/buf.plugin.yaml
+++ b/plugins/bufbuild/connect-query/v0.1.0/buf.plugin.yaml
@@ -1,0 +1,21 @@
+version: v1
+name: buf.build/bufbuild/connect-query
+plugin_version: v0.1.0
+source_url: https://github.com/bufbuild/connect-query
+description: Generates client stubs for calling Connect services with TanStack Query.
+deps:
+  - plugin: buf.build/bufbuild/connect-web:v0.6.0
+output_languages:
+  - javascript
+  - typescript
+registry:
+  npm:
+    import_style: module
+    rewrite_import_path_suffix: connectquery.js
+    deps:
+      - package: '@bufbuild/connect-query'
+        version: ^0.1.0
+      - package: '@bufbuild/protobuf'
+        version: ^1.0.0
+spdx_license_id: Apache-2.0
+license_url: https://github.com/bufbuild/connect-query/blob/main/LICENSE
--- a/plugins/bufbuild/connect-query/source.yaml
+++ b/plugins/bufbuild/connect-query/source.yaml
@@ -1,0 +1,3 @@
+source:
+  npm_registry:
+    name: "@bufbuild/protoc-gen-connect-query"
--- a/plugins/bufbuild/connect-query/v0.1.0/.dockerignore
+++ b/plugins/bufbuild/connect-query/v0.1.0/.dockerignore
@@ -1,0 +1,3 @@
+*
+!Dockerfile
+!package*.json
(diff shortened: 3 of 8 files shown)


plugins/bufbuild/connect-query/v0.1.0/buf.plugin.yaml
```diff
@@ -0,0 +1,21 @@
+version: v1
+name: buf.build/bufbuild/connect-query
+plugin_version: v0.1.0
+source_url: https://github.com/bufbuild/connect-query
+description: Generates client stubs for calling Connect services with TanStack Query.
+deps:
+  - plugin: buf.build/bufbuild/connect-web:v0.6.0
```
pkwarren, line 7: Looking at the package.json for @bufbuild/connect-query, I see it depends on connect-web. This will also pull in protobuf-es (connect-web depends on it).

plugins/bufbuild/connect-query/v0.1.0/buf.plugin.yaml
```diff
@@ -0,0 +1,21 @@
...
+  - typescript
+registry:
+  npm:
+    import_style: module
+    rewrite_import_path_suffix: connectquery.js
+    deps:
+      - package: '@bufbuild/connect-query'
+        version: ^0.1.0
+      - package: '@bufbuild/protobuf'
+        version: ^1.0.0
```
pkwarren, line 19: These were the two runtime dependencies I saw in the generated code.

plugins/bufbuild/connect-query/v0.1.0/buf.plugin.yaml
mfridman, line 7: Does it depend on `connect-web`, or the base types from `protobuf-es`. I have no idea tbh.

pkwarren: You're too speedy - left a comment below. https://github.com/bufbuild/connect-query/blob/160b59f84ddb26126c47ee29ea8eb09ffe5d9d85/packages/connect-query/package.json#L39 shows there is a dependency on connect-web.

mfridman: Hmmm, I would have expected this to depend on `protobuf-es`. You're probably right, but I'd like to double-check if the generated code depends on connect-web generated code or protobuf-es generated code.

I don't understand why it needs the connect-web generated code, but I'm obviously not too familiar with this plugin.

mfridman (approved):
lgtm, mainly a question around the dep in this thread: https://github.com/bufbuild/plugins/pull/330/files#r1088093532

plugins/bufbuild/connect-query/v0.1.0/buf.plugin.yaml
dimitropoulos: Yes, what @paul-sachs said is correct.  The _generated_ code itself does not depend on connect-web, but the functions that are used in the generated code depend on `Transport` at runtime, which is provided by `connect-web`.

Put another way, `connect-web` is a runtime dependency of a TypeScript web app that uses `connect-query`.  There is no way to use `connect-query` without `connect-web` because `connect-query` needs `Transport` from `connect-web` (and nothing else).

paul-sachs: @dimitropoulos I did notice that `connect-query` seems to have a [direct dependency](https://github.com/bufbuild/connect-query/blob/160b59f84ddb26126c47ee29ea8eb09ffe5d9d85/packages/connect-query/package.json#LL41C4-L41C4) on `connect-web` plugin though.  Perhaps that's unintentional?

pkwarren: Updated this plugin to switch the dependency on generated code to bufbuild/es.

dimitropoulos: At runtime, the one and only direct thing used from connect-web is `ConnectError` ([here](https://github.com/bufbuild/connect-query/blob/main/packages/connect-query/src/use-transport.tsx#L16)).  The rest of the direct dependencies are just types for Transport:
Screenshot_20230126_125854

PR #332, opened 2023-01-26
Upload release to GCS bucket

pkwarren:
Update the release workflow to call the upload workflow when a release is completed. Add an upload workflow which downloads (and caches) the latest release and uses `gsutil rsync` to copy its contents to a GCS bucket.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -41,3 +41,6 @@ jobs:
         if: always()
         run: |
           rm -fv minisign.key
+  upload:
+    needs: release
+    uses: ./.github/workflows/upload.yml
--- a/.github/workflows/upload.yml
+++ b/.github/workflows/upload.yml
@@ -1,0 +1,45 @@
+name: Upload to GCS
+
+on:
+  workflow_call:
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+jobs:
+  release:
+    if: github.repository == 'bufbuild/plugins'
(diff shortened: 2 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -41,3 +41,6 @@ jobs:
         if: always()
         run: |
           rm -fv minisign.key
+  upload:
+    needs: release
+    uses: ./.github/workflows/upload.yml
```
pkwarren, line 46: I think this is the syntax to call another workflow but I haven't used this myself before.

.github/workflows/upload.yml
```diff
@@ -0,0 +1,45 @@
+name: Upload to GCS
+
+on:
+  workflow_call:
+  workflow_dispatch:
```
pkwarren, line 5: Made this able to be called manually or from another workflow.

.github/workflows/upload.yml
```diff
@@ -0,0 +1,45 @@
...
+        with:
+          go-version: 1.19.x
+          check-latest: true
+          cache: true
+      - name: Cache Plugins
+        id: cache-plugins
+        uses: actions/cache@v3
+        with:
+          path: downloads
+          key: plugin-downloads
```
pkwarren, line 28: Caching this so we download less stuff each time. Might take this out if the cache uses too much storage, but for right now this should work.

rhbuf (approved):
lgtm

PR #334, opened 2023-01-26
Add id-token permission to upload workflow

pkwarren:


rhbuf (approved):
lgtm - id-token is needed to exchange for gcp credentials

PR #337, opened 2023-01-27
Fix syntax in release workflow

pkwarren:


.github/workflows/release.yml
```diff
@@ -45,5 +45,7 @@ jobs:
           rm -fv minisign.key
   upload:
     environment: production
+    if: github.repository == 'bufbuild/plugins'
+    runs-on: ubuntu-22.04
```
pkwarren, line 49: Hoping this will resolve https://github.com/bufbuild/plugins/actions/runs/4020280875.

rhbuf (approved):
lgtm - matches what is in upload.yml

PR #344, opened 2023-01-31
Download plugin-releases.json.*

pkwarren:
As part of downloading plugins, we should download the corresponding metadata files - plugin-releases.json and plugin-releases.json.minisig.

### DIFF
--- a/internal/release/github.go
+++ b/internal/release/github.go
@@ -151,7 +151,7 @@ func (c *Client) DownloadPluginReleasesToDir(
 		return nil, fmt.Errorf("release %s file %q doesn't match signature %q", release.GetName(), PluginReleasesFile, PluginReleasesSignatureFile)
 	}
 	pluginReleasesDownload := filepath.Join(dir, PluginReleasesFile)
-	if err := os.WriteFile(pluginReleasesDownload, releasesJSONBytes, 0644); err != nil {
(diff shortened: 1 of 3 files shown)


internal/release/github.go
```diff
@@ -129,20 +130,42 @@ func getOwnerRepoFromReleaseURL(url string) (string, string, error) {
...
+	}
+	if err := os.Chtimes(pluginReleasesDownload, releasesJSONLastModified, releasesJSONLastModified); err != nil {
+		return nil, err
+	}
+	if releasesJSONMinisigBytes != nil {
+		pluginReleasesSignatureDownload := filepath.Join(dir, PluginReleasesSignatureFile)
+		if err := os.WriteFile(pluginReleasesSignatureDownload, releasesJSONMinisigBytes, 0644); err != nil {
+			return nil, err
+		}
+		if err := os.Chtimes(pluginReleasesSignatureDownload, releasesJSONMinisigLastModified, releasesJSONMinisigLastModified); err != nil {
```
mfridman, line 165: I don't quite understand why we care about the time component at all here, I think I'm missing something...

mfridman: Ohhh, we're looping over all the assets and then updating the file to whatever the latest release timestamp is. But why does this matter though? Shouldn't this be based on the release name which has a date component..

pkwarren: I'm trying to keep times in sync with the times on the release - it lets me tell at a glance when things changed last.

mfridman (approved):

PR #346, opened 2023-02-01
Attempt to fix caching plugins

pkwarren:
The release workflow isn't noticing when new plugins are updated. We should update the configuration of the cache action to create a new cache when the plugins change.

### DIFF
--- a/.github/workflows/upload.yml
+++ b/.github/workflows/upload.yml
@@ -27,7 +27,9 @@ jobs:
         uses: actions/cache@v3
         with:
           path: downloads
-          key: plugin-downloads
+          key: ${{ runner.os }}-plugin-downloads-${{ hashFiles('**/plugin-releases.json') }}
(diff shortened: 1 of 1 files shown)


.github/workflows/upload.yml
```diff
@@ -27,7 +27,9 @@ jobs:
         uses: actions/cache@v3
         with:
           path: downloads
-          key: plugin-downloads
+          key: ${{ runner.os }}-plugin-downloads-${{ hashFiles('**/plugin-releases.json') }}
```
mfridman, line 30: Curious why any of this needs to be cached, isn't it readily available via a `gh` call to https://github.com/bufbuild/plugins/releases ?

mfridman (approved):
lgtm. 

Although if we could avoid caching and querying this from the source of truth that'd be ideal imo.

PR #367, opened 2023-02-14
Fix source_url in grpc-ecosystem/grpc-gateway

kei2100:
The source_url in buf.plugin.yaml for grpc-gateway seemed to be wrong, so I fixed it.


pkwarren (approved):
Good catch - thanks!

mfridman (approved):

PR #400, opened 2023-03-06
Upgrade dependencies

mfridman:


.github/workflows/ci.yml
```diff
@@ -45,8 +45,9 @@ jobs:
         check-latest: true
         cache: true
     - name: Install buf cli
-      run: |
-        go install github.com/bufbuild/buf/cmd/buf@main
+      uses: bufbuild/buf-setup-action@v1
+      with:
+        version: 1.15.0
```
mfridman, line 50: I suppose we could use `latest` here, given the config has more-or-less stabilized for now.

pkwarren (approved):

.github/workflows/ci.yml
pkwarren: There are a few changes coming - I don't mind keeping this with the specific version. We could however expand the buf-setup-action to target a specific version, so we'll get updates to it automatically (and not need to specify version?).

pkwarren (approved):

PR #401, opened 2023-03-06
Add prost-serde and prost-tonic plugins

### LINKED ISSUE #96: Plugin request for Buf Schema Registry - prost plugins

pkwarren:
Add support for prost-serde and prost-tonic rust plugins. Work around the test failure with insertion points by passing the `no_include` option when running the tests (we currently presume each test can run in isolation). Clean up some older cruft regarding changes to common files - in almost no cases now do we want to rebuild all 202 plugins.

Fixes #96.


.github/workflows/pr.yml
```diff
@@ -28,10 +28,7 @@ jobs:
       with:
         files: |
           plugins/**
-          tests/*.go
           tests/testdata/buf.build/**
-          tests/testdata/images/*.gz
-          Makefile
```
pkwarren, line 34: At one point it made sense to rebuild everything if you changed common files. Now this is way too time consuming for ~200+ plugins. If we ever want to do something like this in the future, maybe we run a weekly full build of all plugins just as a sanity check.

Makefile
```diff
@@ -25,7 +25,7 @@ PLUGIN_IMAGES := $(patsubst %/buf.plugin.yaml,.build/plugin/%/image,$(PLUGIN_YAM
 all: build
 
 .PHONY: build
-build: $(BASE_IMAGES) $(PLUGIN_IMAGES)
```
pkwarren, line 28: This has been long removed.

tests/plugins_test.go
```diff
@@ -46,6 +46,13 @@ exec docker run --log-driver=none --rm -i {{.ImageName}}:{{.Version}} "$@"
 		"eliza",
 		"petapis",
 	}
+	// Options to pass to the plugin during tests. The prost plugins depend on insertion points by default, which
+	// breaks our current test strategy which is to run each plugin in isolation. Override the test options for
+	// these plugins until the tests are updated to support running all plugin dependencies in sequence.
+	testOverrideOptions = map[string][]string{
+		"buf.build/community/neoeinstein-prost-serde": {"no_include"},
+		"buf.build/community/neoeinstein-tonic":       {"no_include"},
+	}
```
pkwarren, line 55: Workaround for now to allow the tests to run for these plugins in isolation.

mfridman (approved):

PR #433, opened 2023-03-20
Add Maven registry metadata to Java/Kotlin plugins

pkwarren:


plugins/bufbuild/connect-kotlin/v0.1.1/buf.plugin.yaml
```diff
@@ -9,3 +9,16 @@ output_languages:
   - kotlin
 spdx_license_id: Apache-2.0
 license_url: https://github.com/bufbuild/connect-kotlin/blob/v0.1.1/LICENSE
+registry:
+  maven:
+    compiler:
+      kotlin:
+        version: 1.8.10
```
pkwarren, line 16: I'd like to validate these in the tests.

plugins/bufbuild/connect-kotlin/v0.1.1/buf.plugin.yaml
```diff
@@ -9,3 +9,16 @@ output_languages:
...
+    deps:
+      - build.buf:connect-kotlin:0.1.1
+      - build.buf:connect-kotlin-google-java-ext:0.1.1
```
pkwarren, line 19: I also think we should validate all deps can be resolved from Maven central.

plugins/protocolbuffers/kotlin/v21.10/buf.plugin.yaml
```diff
@@ -7,3 +7,17 @@ output_languages:
...
 spdx_license_id: BSD-3-Clause
 license_url: https://github.com/protocolbuffers/protobuf/blob/v21.10/LICENSE
+registry:
+  maven:
+    compiler:
+      kotlin:
+        version: 1.8.10
+    deps:
+      - com.google.protobuf:protobuf-kotlin:3.21.10
+      - org.jetbrains.kotlin:kotlin-stdlib-jdk8:1.8.10
```
pkwarren, line 17: Added jdk8 dependency to full runtime, and just normal stdlib to lite runtime. Does that sound reasonable @buildbreaker? I figure anyone using these dependencies can always pull in additional support for newer features.

stefanvanburen (approved):
lgtm, although may want to wait on @buildbreaker's thoughts

PR #435, opened 2023-03-22
Fail fast if plugin execution fails

pkwarren:
Currently, if running the plugin fails for any reason, we continue the tests and fail with an error about the 'gen' directory not existing. We should fail fast to make it more apparent what caused the test failure.


mfridman (approved):
Very reasonable

PR #436, opened 2023-03-22
Add community/scalapb-zio plugin

### LINKED ISSUE #303: Plugin request for Buf Schema Registry - ZIO GRPC

pkwarren:
Fixes #303.


plugins/community/scalapb-zio/v0.5.3/buf.plugin.yaml
```diff
@@ -0,0 +1,9 @@
+version: v1
+name: buf.build/community/scalapb-zio
```
pkwarren, line 2: @11Dimensions - Do you feel this is enough for users to discover the plugin in the BSR? In most cases, we've chose the second part of the community name to be `protoc-gen-{name}`, but in this case I wonder if people will find it as easily as they'd do if we include `grpc` in the name.

@mfridman what do you think?

mfridman: For existing users `protoc-gen-zio` would be sufficient, since they are familiar with the plugin.

For new users, this might not make sense. But I think that's a partially orthogonal problem that's more on us. I.e., the BSR should have metadata like tags that would enable users to filter for scala + grpc to surface the plugin.

I think this is one of few exceptions where `zio-grpc` makes sense. It's also referenced as such in a Stackoverflow questions and the documentation site: https://scalapb.github.io/zio-grpc/

plugins/community/scalapb-zio/v0.5.3/Dockerfile
```diff
@@ -0,0 +1,15 @@
+# syntax=docker/dockerfile:1.4
+FROM debian:bullseye-20230227 AS build
```
mfridman, line 2: Could bump this to `bullseye-20230320`

mfridman (approved):

plugins/community/scalapb-zio/v0.5.3/Dockerfile
pkwarren: Done.

PR #438, opened 2023-03-23
Automate upgrades of Docker base images

pkwarren:
These Dockerfiles will be updated by dependabot automatically and used as input when creating new versions of plugins by the fetch-versions utility.

### DIFF
--- a/.github/docker/Dockerfile.debian
+++ b/.github/docker/Dockerfile.debian
@@ -1,0 +1,3 @@
+FROM debian:bullseye-20221114
+
+CMD echo this is a dummy file used to automate dependency upgrades for plugins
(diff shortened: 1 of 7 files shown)


.github/docker/Dockerfile.debian
```diff
@@ -0,0 +1,3 @@
+FROM debian:bullseye-20221114
```
pkwarren, line 1: Purposefully chose old versions to verify things work.

.github/docker/Dockerfile.golang
```diff
@@ -0,0 +1,3 @@
+FROM golang:1.19.2-bullseye
```
mfridman, line 1: Should this be 1.20.x ?

mfridman: Or is this to test the upgrade path?

pkwarren: To test upgrades.

mfridman: :doh:

... Dockerfiles _will be_ updated by

mfridman (approved):

PR #453, opened 2023-03-28
Simplify plugins build process

pkwarren:
Add support for workflow_dispatch (manual builds) to specify the PLUGINS env var to allow for rebuilding certain plugins (bypassing the default logic that uses changed-files actions to determine what to rebuild).

Update the changed-files action to run just once and print out a value which can be used as the PLUGINS env var. This allows us to isolate this logic around github actions to just a single place in the code.

Update places that decide which plugins to run to be more conservative. If no filter parameters are used, don't build any plugins and show an error message. This should make it better for new contributors to this project to more easily get started with building a specific plugin (since no one should really be building all 300+ plugins at this point unless they really mean to do so).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,7 +22,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - name: set PLUGINS from workflow inputs
-      run: echo "PLUGINS=${inputs.plugins}" >> $GITHUB_ENV
+      if: ${{ inputs.plugins }}
+      run: echo "PLUGINS=${{ inputs.plugins }}" >> $GITHUB_ENV
     - uses: actions/checkout@v3
       with:
         fetch-depth: 0
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,7 +2,12 @@ name: ci
 on:
   push:
     branches: [ "main" ]
-  workflow_dispatch: {} # support manual runs
+  workflow_dispatch:
+    inputs:
+      plugins:
+        description: "Plugins to build and publish (i.e. connect-go:latest, connect-go, grpc/java:v1.53.0)"
+        default: ''
+        type: string
 
 # Ref: https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#permissions
 permissions:
@@ -16,6 +21,8 @@ jobs:
   ci:
     runs-on: ubuntu-latest
     steps:
+    - name: set PLUGINS from workflow inputs
+      run: echo "PLUGINS=${inputs.plugins}" >> $GITHUB_ENV
     - uses: actions/checkout@v3
       with:
         fetch-depth: 0
@@ -44,6 +51,15 @@ jobs:
         go-version: 1.20.x
         check-latest: true
         cache: true
+    - name: Set PLUGINS env var from changed files
+      env:
+        ALL_MODIFIED_FILES: ${{ steps.changed-files.outputs.all_modified_files }}
+        ANY_MODIFIED: ${{ steps.changed-files.outputs.any_modified }}
+      run: |
+        val=`go run ./internal/cmd/changed-plugins .`
+        if [[ -n "${val}" && -z "${PLUGINS}" ]]; then
+          echo "PLUGINS=${val}" >> $GITHUB_ENV
+        fi
     - name: Install buf cli
       uses: bufbuild/buf-setup-action@v1.15.1
       with:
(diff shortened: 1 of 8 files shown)


.github/workflows/ci.yml
```diff
@@ -2,7 +2,12 @@ name: ci
 on:
   push:
     branches: [ "main" ]
-  workflow_dispatch: {} # support manual runs
+  workflow_dispatch:
+    inputs:
+      plugins:
+        description: "Plugins to build and publish (i.e. connect-go:latest, connect-go, grpc/java:v1.53.0)"
+        default: ''
+        type: string
```
pkwarren, line 10: This allows a user to override the calculated changed plugins with a manual build to provide their own inputs on which plugins should be built/tested/published.

.github/workflows/ci.yml
```diff
@@ -44,6 +51,15 @@ jobs:
...
         cache: true
+    - name: Set PLUGINS env var from changed files
+      env:
+        ALL_MODIFIED_FILES: ${{ steps.changed-files.outputs.all_modified_files }}
+        ANY_MODIFIED: ${{ steps.changed-files.outputs.any_modified }}
+      run: |
+        val=`go run ./internal/cmd/changed-plugins .`
+        if [[ -n "${val}" && -z "${PLUGINS}" ]]; then
+          echo "PLUGINS=${val}" >> $GITHUB_ENV
+        fi
```
pkwarren, line 62: This runs the new utility that uses the github actions variables to determine which plugins have changed. This only runs if someone didn't manually set PLUGINS to a valid value above.

.github/workflows/ci.yml
```diff
@@ -69,14 +85,8 @@ jobs:
...
     - name: Build
-      env:
-        ALL_MODIFIED_FILES: ${{ steps.changed-files.outputs.all_modified_files }}
-        ANY_MODIFIED: ${{ steps.changed-files.outputs.any_modified }}
       shell: bash
       run: make DOCKER_ORG="ghcr.io/bufbuild" DOCKER_BUILD_EXTRA_ARGS="--push"
     - name: Test
-      env:
-        ALL_MODIFIED_FILES: ${{ steps.changed-files.outputs.all_modified_files }}
-        ANY_MODIFIED: ${{ steps.changed-files.outputs.any_modified }}
```
pkwarren, line 80: All of these commands will just use one set of inputs - the `PLUGINS` env var.

Makefile
```diff
@@ -23,6 +23,11 @@ PLUGIN_IMAGES := $(patsubst %/buf.plugin.yaml,.build/plugin/%/image,$(PLUGIN_YAM
 
 .PHONY: all
 all: build
+	@if [[ -z "${PLUGINS}" ]]; then \
+		echo "No plugins specified to build with PLUGINS env var."; \
+		echo "See Makefile for example PLUGINS env var usage."; \
+		echo "To build all plugins (will take a long time), build with 'make PLUGINS=all'."; \
+	fi
```
pkwarren, line 30: Prints a nicer message if someone runs make when starting out with this project and didn't choose the plugins to build.

cmd/dependency-order/main.go
```diff
@@ -27,19 +25,9 @@ func main() {
...
-		if err != nil {
-			log.Fatalf("failed to filter plugins by PLUGINS env var: %v", err)
-		}
-	} else {
-		// Filter by changed plugins (for PR builds)
-		includedPlugins, err = plugin.FilterByChangedFiles(plugins, envconfig.OsLookuper())
-		if err != nil {
-			log.Fatalf("failed to filter plugins by changed files: %v", err)
-		}
+	includedPlugins, err := plugin.FilterByPluginsEnv(plugins, os.Getenv("PLUGINS"))
```
pkwarren, line 28: Now all commands other than changed-plugins only look at one source of information instead of duplicating this logic across them all.

internal/plugin/plugin.go
```diff
@@ -147,6 +147,9 @@ func Load(path string, basedir string) (*Plugin, error) {
 // FilterByPluginsEnv returns matching plugins based on a space separated list of plugins (and optional versions) to include.
 func FilterByPluginsEnv(plugins []*Plugin, pluginsEnv string) ([]*Plugin, error) {
 	if pluginsEnv == "" {
+		return nil, nil
+	}
```
pkwarren, line 151: Made these utilities more strict so if no filter is provided, they will default to not building anything.

twoism (approved):

internal/cmd/changed-plugins/main.go
```diff
@@ -0,0 +1,43 @@
...
+		log.Fatalf("failed to filter plugins by changed files: %v", err)
+	}
+	var sb strings.Builder
+	for _, includedPlugin := range includedPlugins {
+		sb.WriteString(strings.TrimPrefix(includedPlugin.Name, "buf.build/"))
+		sb.WriteByte(':')
+		sb.WriteString(includedPlugin.PluginVersion)
+		sb.WriteByte(' ')
+	}
+	fmt.Println(strings.TrimSpace(sb.String())) //nolint:forbidigo
```
pkwarren, line 42: Sample run:

```
$ ANY_MODIFIED=true \
ALL_MODIFIED_FILES=plugins/community/stephenh-ts-proto/v1.145.0/.dockerignore,plugins/community/stephenh-ts-proto/v1.145.0/Dockerfile,plugins/community/stephenh-ts-proto/v1.145.0/buf.plugin.yaml,plugins/community/stephenh-ts-proto/v1.145.0/package-lock.json,plugins/community/stephenh-ts-proto/v1.145.0/package.json \
go run ./internal/cmd/changed-plugins .
2023/03/28 17:56:31 including plugin: plugins/community/stephenh-ts-proto/v1.145.0/buf.plugin.yaml
community/stephenh-ts-proto:v1.145.0
```

mfridman (approved):
Looks great 🎉.

PR #456, opened 2023-03-29
Update Docker plugins build

pkwarren:
Add support for running Docker builds in parallel (with knowledge of dependencies between plugins to add further speed ups). Remove complexity of builds from Makefile to a Go CLI that can more easily be used (and tested) when adding wasm plugins.


.github/workflows/ci.yml
```diff
@@ -85,9 +88,8 @@ jobs:
         registry: ghcr.io
         username: ${{ github.actor }}
         password: ${{ secrets.GITHUB_TOKEN }}
-    - name: Build
-      shell: bash
-      run: make DOCKER_ORG="ghcr.io/bufbuild" DOCKER_BUILD_EXTRA_ARGS="--push"
```
pkwarren, line 90: This was one thing I wanted to fix from before - it was pushing the images to the registry prior to testing. I want the tests to pass before we push anything.

.github/workflows/fetch_versions.yml
```diff
@@ -37,7 +37,7 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ steps.generate_token.outputs.token }}
         run: |
-          go run ./cmd/fetcher .
+          go run ./internal/cmd/fetcher .
```
pkwarren, line 40: Relocated all utilities which have no real use outside this project to internal.

internal/cmd/dockerbuild/main.go
```diff
@@ -0,0 +1,219 @@
...
+)
+
+// dockerbuild is a helper program used to build plugins from Dockerfiles in an optimized fashion.
+// It replaces some clunky (and non-parallel) jobs in the Makefile.
+// It knows about relationships between common containers (like protoc/grpc Bazel plugins),
+// which enables optimized builds which build common code first before binaries.
+// It also makes it easier to add new labels to images using existing code to parse buf.plugin.yaml.
+
+// larger amount of parallelism lead to OOM errors in testing - clamp for now
+const maxLimit = 8
```
pkwarren, line 30: If I let this run with 10 cores (m1) I'd see random times where the process would get killed trying to build all images. Dialed this down smaller to hopefully avoid any memory OOM killer.

internal/cmd/dockerbuild/main.go
```diff
@@ -0,0 +1,219 @@
...
+			ctx,
+			dockerCmd,
+			commonArgs...,
+		)
+		cmd.Args = append(
+			cmd.Args,
+			"--target",
+			stage,
+			"-t",
+			fmt.Sprintf("%s/plugins-%s-%s:%s-%s", dockerOrg, identity.Owner(), identity.Plugin(), plugin.PluginVersion, stage),
```
pkwarren, line 166: By saving results of each build stage, we can ensure when building multiple versions of plugins like protoc/grpc ones, we'll save the build result from prior builds. Otherwise, each version build would stomp on each other.

internal/cmd/dockerbuild/main.go
```diff
@@ -0,0 +1,219 @@
...
+	cmd.Args = append(
+		cmd.Args,
+		"-t",
+		fmt.Sprintf("%s/plugins-%s-%s:%s", dockerOrg, identity.Owner(), identity.Plugin(), plugin.PluginVersion),
+	)
+	cmd.Args = append(cmd.Args, filepath.Dir(plugin.Path))
+	return cmd.CombinedOutput()
+}
+
+func getDockerBuildStages(dockerfile string) (_ []string, retErr error) {
```
pkwarren, line 189: Parses Docker stage names from a multi-stage build.

PR #457, opened 2023-03-29
Ignore changed-files calculation for manual build

pkwarren:
Update the GitHub action workflows to ignore use of changed-files action if we've already gotten a non-empty input to the workflow.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -33,11 +33,13 @@ jobs:
         fetch-depth: 0
     - uses: nrwl/nx-set-shas@177b48373c6dc583ce0d9257ffb484bdd232fedf
       id: last_successful_commit_push
+      if: ${{ inputs.plugins != '' }}
       with:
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -33,11 +33,13 @@ jobs:
         fetch-depth: 0
     - uses: nrwl/nx-set-shas@177b48373c6dc583ce0d9257ffb484bdd232fedf
       id: last_successful_commit_push
+      if: ${{ inputs.plugins != '' }}
       with:
         main-branch-name: ${{ github.ref_name }}
         workflow-id: 'ci.yml'
     - name: Get changed files
       id: changed-files
+      if: ${{ inputs.plugins != '' }}
```
pkwarren, line 42: Tried kicking off a manual build to rebuild an individual plugin and got https://github.com/bufbuild/plugins/actions/runs/4557846013/jobs/8040014496. This should skip all those steps if we've set the workflow variable already.

PR #459, opened 2023-03-29
Add dockerpush command to run after dockerbuild

pkwarren:
Currently, we're re-running the docker build with the '--push' argument, which unfortunately both triggers a rebuild and pushes the images for the other stages of the Dockerfile (which can be much larger than the actual plugin image).

Add a new utility `dockerpush` to run after `dockerbuild` which only does the push of pre-built images. This should speed up the build and also reduce what we push to GHCR.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,6 +9,9 @@ on:
         default: ''
         type: string
 
+env:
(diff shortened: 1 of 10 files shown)


twoism (approved):
Nice!

PR #471, opened 2023-04-04
Update cron intervals

mfridman:
I suggest we change the Fetcher to run in the morning 10AM eastern, that way we get a notification to remind us on the same day (instead of 8PM the night before).

And update the Release workflow to run every hour, so from the time we merge the PR it gets release. Although maybe this one we can reduce to 4 hours, no strong preference.

### DIFF
--- a/.github/workflows/fetch_versions.yml
+++ b/.github/workflows/fetch_versions.yml
@@ -2,7 +2,8 @@ name: Fetch latest versions
 
 on:
   schedule:
-    - cron: "0 0 * * *" # Run once a day
+    # Run once a day at 10 AM EST
+    - cron: "0 14 * * *"
   workflow_dispatch:
 
 jobs:
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -2,7 +2,8 @@ name: Create Release
 
 on:
   schedule:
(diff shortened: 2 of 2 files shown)


.github/workflows/fetch_versions.yml
```diff
@@ -2,7 +2,8 @@ name: Fetch latest versions
 
 on:
   schedule:
-    - cron: "0 0 * * *" # Run once a day
+    # Run once a day at 10 AM EST
+    - cron: "0 14 * * *"
```
pkwarren, line 6: Nice - this will be much better.

.github/workflows/release.yml
```diff
@@ -2,7 +2,8 @@ name: Create Release
 
 on:
   schedule:
-    - cron: "0 0 * * *" # Run once a day
+    # Run every hour.
+    - cron: "30 * * * *"
```
pkwarren, line 6: I'd rather we leave ourselves some leeway here - in one case because of a bad github action we had to revert everything prior to the release running. Maybe drop down to 4/8 hours to start.

mfridman: Let's meet in the middle, 6 hours .. which is ~4 times per day?

pkwarren (approved):

PR #486, opened 2023-04-12
Prepend buf.build/ when matching plugins

pkwarren:
The plugin-releases.json file only has the "{owner}/{plugin}", not the leading "{remote}". For the purposes of matching, we should include the "{remote}/" prefix for consistency.

### DIFF
--- a/cmd/download-plugins/main.go
+++ b/cmd/download-plugins/main.go
@@ -93,9 +93,14 @@ func run() error {
 		// Filter out plugins which aren't specified in PLUGINS env var
 		if includePlugins != nil {
 			var matched bool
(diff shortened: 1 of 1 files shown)


cmd/download-plugins/main.go
```diff
@@ -93,9 +93,14 @@ func run() error {
 		// Filter out plugins which aren't specified in PLUGINS env var
 		if includePlugins != nil {
 			var matched bool
+			// For the purposes of matching, we should include the full name as defined in buf.plugin.yaml
+			pluginName := pluginRelease.PluginName
+			if !strings.HasPrefix(pluginName, "buf.build/") {
+				pluginName = "buf.build/" + pluginName
```
mfridman, line 99: Oof

PR #491, opened 2023-04-14
Kotlin generated code depends on Java code

pkwarren:


plugins/bufbuild/connect-kotlin/v0.1.4/buf.plugin.yaml
```diff
@@ -4,7 +4,7 @@ plugin_version: v0.1.4
 source_url: https://github.com/bufbuild/connect-kotlin
 description: Idiomatic gRPC & Connect RPCs for Kotlin.
 deps:
-  - plugin: buf.build/protocolbuffers/java:v22.2
+  - plugin: buf.build/protocolbuffers/kotlin:v22.2
```
buildbreaker, line 7: Just to make sure: the transitive dependency with the java plugin gets applied here, right?

buildbreaker (approved):

plugins/bufbuild/connect-kotlin/v0.1.4/buf.plugin.yaml
pkwarren: Correct.

PR #499, opened 2023-04-20
Improve PLUGINS var handling and validate names

pkwarren:
Improve the handling of the PLUGINS variable to allow either space or comma separated plugin(s). Update name validation to prevent use of underscores which will cause issues with remote packages in some languages.

### DIFF
--- a/internal/plugin/plugin.go
+++ b/internal/plugin/plugin.go
@@ -10,6 +10,7 @@ import (
 	"sort"
 	"strconv"
 	"strings"
+	"unicode"
 
 	"github.com/bufbuild/buf/private/bufpkg/bufplugin/bufpluginconfig"
 	"github.com/bufbuild/buf/private/pkg/encoding"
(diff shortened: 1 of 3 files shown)


internal/plugin/plugin.go
```diff
@@ -230,7 +231,9 @@ func FilterByChangedFiles(plugins []*Plugin, lookuper envconfig.Lookuper) ([]*Pl
 
 func ParsePluginsEnvVar(pluginsEnv string) ([]IncludePlugin, error) {
 	var includes []IncludePlugin
-	fields := strings.Fields(pluginsEnv)
+	fields := strings.FieldsFunc(pluginsEnv, func(r rune) bool {
+		return unicode.IsSpace(r) || r == ','
```
pkwarren, line 235: This has tripped me up a few times and space delimited values is a pain to work with sometimes in the shell so also accept comma delimited plugins.

tests/plugins_test.go
```diff
@@ -139,6 +139,8 @@ func TestBufPluginConfig(t *testing.T) {
 		assert.NotEmpty(t, config.PluginVersion)
 		assert.NotEmpty(t, config.SPDXLicenseID)
 		assert.NotEmpty(t, config.LicenseURL)
+		// Don't allow underscore in plugin names - this would cause issues in remote packages
+		assert.NotContains(t, config.Name.IdentityString(), "_")
```
pkwarren, line 143: In some registries this can cause an issue. To fail fast when we are adding new plugins instead of later at validation time, validate the plugin name here.

mfridman (approved):

PR #508, opened 2023-04-28
Add Swift registry config to swift plugins

pgmitche:


pgmitche:
It looks like I also need to get a new release of the buf CLI that includes the registry changes before I can merge this as well

pkwarren (approved):
Changes look fine to me. I'd merge when you're pretty sure the registry config won't change any further.

pkwarren (approved):

plugins/apple/swift/v1.21.0/Dockerfile
```diff
@@ -4,6 +4,62 @@ FROM swift:5.7.3-focal AS build
 WORKDIR /app
 RUN git clone --depth 1 --branch 1.21.0 https://github.com/apple/swift-protobuf --recursive
 WORKDIR /app/swift-protobuf
+RUN git apply <<'EOF'
```
pgmitche, line 7: Patches in functionality for the `ExtraModuleImports` option just like 
https://github.com/grpc/grpc-swift/blob/75b390e901c7d70af9b4c5ca2677e035f00cd1ab/Sources/protoc-gen-grpc-swift/options.swift#L125
https://github.com/bufbuild/connect-swift/blob/15387ccdde0d6f749872fa1c84776c6816bc9f60/Plugins/ConnectPluginUtilities/GeneratorOptions.swift#L109

pkwarren: This was the pattern we used to use in these builds since we'd build the Docker images with just the Dockerfile as context. Going forward, let's create a .patch file and ADD it like you'd normally do (and update .dockerignore to allow it).

Is this something that we're looking to propose as an upstream patch too?

pgmitche: > Is this something that we're looking to propose as an upstream patch too?

@pkwarren yes, given the grpc/swift precedent here I'm hopeful that it wouldn't take very long to get it accepted as a change in the apple/swift-protobuf repo. So definitely going to give that a go

pkwarren (approved):

pkwarren (approved):

PR #514, opened 2023-05-04
Knit TypeScript plugin

srikrsna-buf:
The TypeScript plugin for Knit.


plugins/bufbuild/knit-ts/v0.0.1/Dockerfile
```diff
@@ -0,0 +1,11 @@
+# syntax=docker/dockerfile:1.4
+FROM node:18.15.0-alpine3.17 AS build
```
mfridman, line 2:
```suggestion
FROM node:18.16.0-alpine3.17 AS build
```

plugins/bufbuild/knit-ts/v0.0.1/buf.plugin.yaml
```diff
@@ -0,0 +1,13 @@
+version: v1
+name: buf.build/bufbuild/knit-ts
+plugin_version: v0.0.1
+source_url: https://github.com/bufbuild/knit-ts
+description: Generates Knit schemas needed to use the Knit client in web browsers and Node.js.
+output_languages:
+  - typescript
+registry:
+  npm:
```
mfridman, line 9: Does it have any runtime `deps`?

srikrsna-buf: None

mfridman (approved):

plugins/bufbuild/knit-ts/v0.0.1/buf.plugin.yaml
pkwarren: The demo depends on `@bufbuild/knit` - is that not needed?

https://github.com/bufbuild/knit-demo/blob/8766e187d5730da22783a22c65a398cbd90d7f21/ts/package.json#L13-L15

srikrsna-buf: The generated code doesn't import it, this should be only if the generated code uses it right?

pkwarren: Yep.

PR #537, opened 2023-05-18
Use buildx --load command

pkwarren:
In order to test against Docker images, we need to save them to the local Docker host.

### DIFF
--- a/internal/docker/build.go
+++ b/internal/docker/build.go
@@ -37,6 +37,7 @@ func Build(
 	commonArgs := []string{
 		"buildx",
 		"build",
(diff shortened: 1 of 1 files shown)


internal/docker/build.go
```diff
@@ -37,6 +37,7 @@ func Build(
 	commonArgs := []string{
 		"buildx",
 		"build",
+		"--load",
```
mfridman, line 40: If this doesn't work, let's revert to what was working and not spend more time on this?

mfridman (approved):

PR #544, opened 2023-05-23
Update grpc/kotlin to newer dependencies

pkwarren:
The grpc/kotlin plugin currently doesn't work with remote packages, as it was dependent on a version of protocolbuffers/kotlin that didn't include a fix for #491. Update this plugin to depend on newer dependencies so that it can work with remote packages.

### DIFF
--- a/plugins/grpc/kotlin/v1.3.0/buf.plugin.yaml
+++ b/plugins/grpc/kotlin/v1.3.0/buf.plugin.yaml
@@ -4,8 +4,8 @@ plugin_version: v1.3.0
 source_url: https://github.com/grpc/grpc-kotlin
 description: Generates Kotlin client and server stubs for the gRPC framework.
 deps:
-  - plugin: buf.build/grpc/java:v1.54.0
-  - plugin: buf.build/protocolbuffers/kotlin:v22.2
+  - plugin: buf.build/grpc/java:v1.54.1
+  - plugin: buf.build/protocolbuffers/kotlin:v22.3
 output_languages:
   - kotlin
 spdx_license_id: Apache-2.0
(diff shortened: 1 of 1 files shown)


plugins/grpc/kotlin/v1.3.0/buf.plugin.yaml
```diff
@@ -14,7 +14,7 @@ registry:
   maven:
     compiler:
       kotlin:
-        version: 1.8.10
+        version: 1.8.21
```
pkwarren, line 17: Went ahead and bumped this to the latest version while we're here.

plugins/grpc/kotlin/v1.3.0/buf.plugin.yaml
```diff
@@ -4,8 +4,8 @@ plugin_version: v1.3.0
 source_url: https://github.com/grpc/grpc-kotlin
 description: Generates Kotlin client and server stubs for the gRPC framework.
 deps:
-  - plugin: buf.build/grpc/java:v1.54.0
-  - plugin: buf.build/protocolbuffers/kotlin:v22.2
+  - plugin: buf.build/grpc/java:v1.54.1
```
mfridman, line 7: Why isn't grpc/java latest version `v1.55.1 and protocolbuffers/kotlin version `v23.1` ?

There's probably a reason, should we add a comment for those not aware?

pkwarren: For now - just doing incremental changes to the levels of other plugins (like connect-kotlin), but if this plugin is ever updated, we can definitely bump further.

mfridman: Yep, sounds perfectly reasonable!

PR #548, opened 2023-05-26
Patch grpc gateway to be remote packages compatible

pgmitche:
Returns @pkwarren's gRPC file changes to the grpc-gateway patch. 
Adds a special case to combat the injected `go_package` option overwriting the valid path for `buf.build/gen/go/..../grpc/go` to `buf.build/gen/go/.../protocolbuffers/go`
Includes plugin registry config changes to enable remote package support


pgmitche:
@pkwarren @mfridman if you guy's get any time this week to have a look it'd be good to close this out, or apply your feedback 🙏🏼

plugins/grpc-ecosystem/gateway/v2.15.2/Dockerfile
```diff
@@ -1,11 +1,18 @@
 # syntax=docker/dockerfile:1.4
 FROM golang:1.20.1-bullseye AS build
```
pkwarren, line 2: While we're here updating the Dockerfile, we can bump this to 1.20.4.

plugins/grpc-ecosystem/gateway/v2.15.2/Dockerfile
```diff
@@ -1,11 +1,18 @@
...
+
+COPY separate_pkg_additional_imports.patch /tmp/separate_pkg_additional_imports.patch
```
pkwarren, line 4:
```suggestion
COPY --link separate_pkg_additional_imports.patch /tmp/separate_pkg_additional_imports.patch
```

I think it is also worth adding this after the git clone so that layer will be cached more often if we need to tweak the patch.

plugins/grpc-ecosystem/gateway/v2.15.2/buf.plugin.yaml
```diff
@@ -5,6 +5,15 @@ source_url: https://github.com/grpc-ecosystem/grpc-gateway
...
 output_languages:
   - go
+registry:
+  go:
+    deps:
+      - module: github.com/grpc-ecosystem/grpc-gateway/v2
+        version: v2.15.2
+  opts:
+    - paths=source_relative
+    - standalone=true
```
pkwarren, line 15: Can you refresh my memory - why is this needed here?

pkwarren (commented):
Few initial comments - still need to dig in more on the patch.

pgmitche:
@pkwarren the output of the remote plugin + grpc-ecosystem/grpc-gateway module

[v2@v2.15.2-20230530153543-ac2725d1cf55.2.zip](https://github.com/bufbuild/plugins/files/11601926/v2%40v2.15.2-20230530153543-ac2725d1cf55.2.zip)

plugins/grpc-ecosystem/gateway/v2.15.2/separate_pkg_additional_imports.patch
```diff
@@ -0,0 +1,291 @@
...
++		if r.separatePackage {
++			// when generating a separate package for the gateway, we need to generate an import statement
++			// for the gRPC stubs that are no longer in the same package. This is done by adding the grpc
++			// package to the additionalImports list. In order to prepare a valid import statement, we'll replace
++			// the source package name, something like: ../pet/v1/v1petgateway with ../pet/v1/v1petgrpc
++			const (
++				baseTypePackageName = "protocolbuffers"
++				grpcPackageName     = "grpc"
++			)
++			packageName := strings.ReplaceAll(svc.File.GoPkg.Name, "gateway", "") + grpcPackageName
```
pkwarren, line 79: I'm worried about doing `ReplaceAll` here - don't we just want to trim the `gateway` suffix here and then append grpc?

plugins/grpc-ecosystem/gateway/v2.15.2/separate_pkg_additional_imports.patch
```diff
@@ -0,0 +1,291 @@
...
++			svc.GRPCFile = &File{
++				GoPkg: GoPackage{
++					// additionally, as the `go_package` option is passed through from the generator, and can only be
++					// set the one time, without making major changes, we'll use the package name sent through the
++					// options as a basis, and replace the source package name with the grpc package name.
++					Path: strings.ReplaceAll(
++						filepath.Join(svc.File.GoPkg.Path, packageName),
```
pkwarren, line 86: Similarly - this shouldn't use ReplaceAll - I think we want to replace `/gen/go/protocolbuffers/go/` with `/gen/go/grpc/go/` here or something more specific.

plugins/grpc-ecosystem/gateway/v2.15.2/separate_pkg_additional_imports.patch
```diff
@@ -0,0 +1,291 @@
...
+ 	Methods []*Method
+ 	// ForcePrefixedName when set to true, prefixes a type with a package prefix.
+@@ -173,10 +176,11 @@ type Service struct {
+ // FQSN returns the fully qualified service name of this service.
+ func (s *Service) FQSN() string {
+ 	components := []string{""}
+-	if s.File.Package != nil {
+-		components = append(components, s.File.GetPackage())
++	if s.GRPCFile != nil && s.GRPCFile.GetPackage() != "" {
++		components = append(components)
```
pkwarren, line 119: Do you need an argument here to the append?

pkwarren (commented):
A few comments on the changes. It would be great if we could add some tests for this change since it is a larger patch than other ones.

plugins/grpc-ecosystem/gateway/v2.15.2/separate_pkg_additional_imports.patch
pgmitche: You're right, the only caveat is that the module name is placed between `/gen/go` & `/protocolbuffers/go`.
I'll update this to be more specific

pgmitche:
> A few comments on the changes. It would be great if we could add some tests for this change since it is a larger patch than other ones.

@pkwarren with regard to tests. Do you mean that you'd like to see the patch diff contain tests that validate the output import paths for example?

plugins/grpc-ecosystem/gateway/v2.15.2/separate_pkg_additional_imports.patch
```diff
@@ -0,0 +1,384 @@
...
++	// Require the two dependencies to be declared as imported packages
++	for _, expectedImport := range []string{
++		`extexamplepb "example.com/path/to/example"`,
++		`"example.com/path/to/example/examplepbgrpc"`,
++	} {
++		if !strings.Contains(result[0].GetContent(), expectedImport) {
++			t.Fatalf("expected to find import %q in the generated file", expectedImport)
++		}
++	}
++}
```
pgmitche, line 326: @pkwarren relevant test asserions

tests/plugins_test.go
```diff
@@ -44,6 +44,7 @@ exec docker run --log-driver=none --rm -i {{.ImageName}}:{{.Version}} "$@"
 	images = []string{
 		"eliza",
 		"petapis",
+		"grpc-gateway",
```
pkwarren, line 47: By adding this here I think it will require plugin.sum files to be generated for this input for all existing plugins. I'd rather we only test this input when testing the gateway plugin.

pgmitche: Ahh sorry, yes let me fix that up. You did already mention that, i think

pkwarren (approved):
Great work!

PR #559, opened 2023-05-31
Update makefile

mfridman:
The `buf` command for uploading plugins has been updated.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,7 +8,7 @@ DOCKER_BUILD_EXTRA_ARGS ?=
 GO_TEST_FLAGS ?= -race -count=1
 
 BUF ?= buf
-BUF_PLUGIN_PUSH_ARGS ?=
+BUF_PLUGIN_PUSH_ARGS ?= --visibility=public
 
 # Specify a space separated list of plugin name (and optional version) to just build/test individual plugins.
 # For example:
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -51,5 +51,5 @@ push: build
 		if [[ "$(DOCKER_ORG)" = "ghcr.io/bufbuild" ]]; then \
 			$(DOCKER) pull $(DOCKER_ORG)/plugins-$${PLUGIN_OWNER}-$${PLUGIN_NAME}:$${PLUGIN_VERSION} || exit 1; \
 		fi; \
-		$(BUF) alpha plugin push $${plugin_dir} $(BUF_PLUGIN_PUSH_ARGS) --image $(DOCKER_ORG)/plugins-$${PLUGIN_OWNER}-$${PLUGIN_NAME}:$${PLUGIN_VERSION} || exit 1; \
+		$(BUF) beta registry plugin push $${plugin_dir} $(BUF_PLUGIN_PUSH_ARGS) --image $(DOCKER_ORG)/plugins-$${PLUGIN_OWNER}-$${PLUGIN_NAME}:$${PLUGIN_VERSION} || exit 1; \
```
pkwarren, line 54: Should we also update the version of buf in go.mod? I know we aren't calling this elsewhere in the CI builds any longer so don't need to update those.

pkwarren (approved):

pkwarren (approved):

Makefile
mfridman: I think this uses the `buf` version that's currently installed, and internally we'll have this set to always be the latest. (I'm updating that next to make sure certain workflows continue to work).

PR #572, opened 2023-06-05
Revert remote packages support for grpc-gateway

pkwarren:
The patch is causing some issues with import aliases when used as a remote plugin. Update to remove the patch and disable remote packages support until we can debug further.


mfridman (approved):
Thanks for reverting this.

PR #574, opened 2023-06-05
Add swift v1.22.0 with updated patch

pkwarren:
Update the patch to cleanly apply against apple/swift v1.22.0.

### DIFF
--- a/plugins/apple/swift/v1.22.0/extramoduleimports.patch
+++ b/plugins/apple/swift/v1.22.0/extramoduleimports.patch
@@ -1,0 +1,58 @@
+diff --git a/Sources/protoc-gen-swift/FileGenerator.swift b/Sources/protoc-gen-swift/FileGenerator.swift
+index 90fd06a6..93460574 100644
+--- a/Sources/protoc-gen-swift/FileGenerator.swift
++++ b/Sources/protoc-gen-swift/FileGenerator.swift
+@@ -121,6 +121,13 @@ class FileGenerator {
+                 p.print("\(visibilityAnnotation)import \(i)\n")
+             }
+         }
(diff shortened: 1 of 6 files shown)


plugins/apple/swift/v1.22.0/extramoduleimports.patch
```diff
@@ -0,0 +1,58 @@
+diff --git a/Sources/protoc-gen-swift/FileGenerator.swift b/Sources/protoc-gen-swift/FileGenerator.swift
```
pkwarren, line 1: Updated the patch to apply cleanly on this version - there were some changes upstream in https://github.com/apple/swift-protobuf/commit/6b33e90eb5655de83a5716801e50d3b6925a5b49 that caused the previous patch not to apply.

@pgmitche / @stefanvanburen / @rebello95 - I'm not sure if we want to do anything here with the new visibilityAnnotation logic?

pgmitche: The patch looks essentially identical, so happy to approve on that basis. 

If I'm understanding the annotation correctly, then it sounds like symbols in the packages generated with that annotation would not be importable, so wouldn't be compatible with remote packages.

pgmitche (approved):

plugins/apple/swift/v1.22.0/extramoduleimports.patch
rebello95: Don't think we need to do anything for that new option, we don't have a use case for implementation-only imports with remote packages. Also updating Connect-Swift: https://github.com/bufbuild/connect-swift/pull/136

PR #606, opened 2023-06-20
Automate bazel upgrades

pkwarren:
Now that we have a Dockerfile.bazel that is kept up to date with dependabot, we can automate upgrades for packages that use Bazel.


mfridman (approved):
hehe, nice!

PR #609, opened 2023-06-21
Add neoeinstein-prost-crate v0.3.1 plugin

### LINKED ISSUE #608: Plugin request for Buf Schema Registry: protoc-gen-prost-crate
Rust plugins were previously added to this repository [here](https://github.com/bufbuild/plugins/issues/96) but protoc-gen-prost-crate was not.

Can protoc-gen-prost-crate be added so that we can move away from using remote generation alpha.

We are currently using the remote plugin with `no_features` to generate a `mod.rs` file in the output directory and a static `Cargo.toml`.

```yaml
  - remote: buf.build/prost/plugins/crate:v0.3.1-1
    out: autogen/lang/rust
    opt:
      - no_features
```

## Mandatory

https://github.com/neoeinstein/protoc-gen-prost/tree/main/protoc-gen-prost-crate

## Optional

https://github.com/neoeinstein/protoc-gen-prost/releases/tag/protoc-gen-prost-crate-v0.2.0

https://buf.build/community/neoeinstein-prost

cobbinma:
add neoeinstein-prost-crate to community plugins

Closes #608

### DIFF
--- a/plugins/community/neoeinstein-prost-crate/v0.3.1/Dockerfile
+++ b/plugins/community/neoeinstein-prost-crate/v0.3.1/Dockerfile
@@ -1,0 +1,12 @@
+# syntax=docker/dockerfile:1.4
+FROM rust:1.68.0-alpine3.17 as builder
+RUN apk add --no-cache musl-dev
+WORKDIR /app
+ENV CARGO_REGISTRIES_CRATES_IO_PROTOCOL=sparse
+RUN --mount=type=cache,target=/usr/local/cargo/registry --mount=type=cache,target=/root/target \
+    cargo install protoc-gen-prost-crate --version 0.3.1 --locked --root /app
+
+FROM gcr.io/distroless/static
+COPY --from=builder /app/bin/protoc-gen-prost-crate /protoc-gen-prost-crate
+USER nobody
+ENTRYPOINT ["/protoc-gen-prost-crate"]
--- a/plugins/community/neoeinstein-prost-crate/v0.3.1/Dockerfile
+++ b/plugins/community/neoeinstein-prost-crate/v0.3.1/Dockerfile
@@ -1,5 +1,5 @@
 # syntax=docker/dockerfile:1.4
-FROM rust:1.68.0-alpine3.17 as builder
+FROM rust:1.70.0-alpine3.18 as builder
 RUN apk add --no-cache musl-dev
 WORKDIR /app
 ENV CARGO_REGISTRIES_CRATES_IO_PROTOCOL=sparse
--- a/tests/plugins_test.go
+++ b/tests/plugins_test.go
@@ -55,6 +55,7 @@ exec docker run --log-driver=none --rm -i {{.ImageName}}:{{.Version}} "$@"
 	// these plugins until the tests are updated to support running all plugin dependencies in sequence.
 	testOverrideOptions = map[string][]string{
 		"buf.build/community/neoeinstein-prost-serde": {"no_include"},
+		"buf.build/community/neoeinstein-prost-crate": {"no_features"},
 		"buf.build/community/neoeinstein-tonic":       {"no_include"},
 	}
 )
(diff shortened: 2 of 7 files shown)


pkwarren:
Thanks @cobbinma for the contribution. We previously discussed the prost-crate plugin here: https://github.com/bufbuild/plugins/issues/96#issuecomment-1338007883

It relies on accessing files which aren't available to remote plugins, so it would need to be used with `no_features` to work as a remote plugin. At the time of getting the other prost plugins working, we opted to skip this one as it most likely won't work as expected. In order to gain full control over all of the plugin's features, you'd want to execute it as a local plugin as described here: https://github.com/neoeinstein/protoc-gen-prost/tree/main/protoc-gen-prost-crate#usage-with-buf

cobbinma:
Thanks @pkwarren.

As I only need the features that can be run remotely, can the plugin be added so that we can use it with `no_features`?

I am currently using it this way and would like to keep doing so:
```yaml
  - remote: buf.build/prost/plugins/crate:v0.3.1-1
    out: autogen/lang/rust
    opt:
      - no_features
```

mfridman:
> As I only need the features that can be run remotely, can the plugin be added so that we can use it with no_features?

Although this works for some use cases, it's also not something we can easily test. I.e., all plugins in this repository are run through at least 1 module and the output is committed as a sum file. We would have to special case this plugin and trust it "just works" without any validation.

Furthermore, `no_features` is not the default, so if users are running this plugin remotely it'll fail and lead to surprising behaviour. This leads to a bit of a maintenance burden for us.

I'm hesitant to add this plugin unless it has sane default behaviour and can be tested.

cobbinma:
@mfridman I have added go sum files and a plugins test override, similar to the other rust plugins.

I agree with your comments and think it is unfortunate `no_features` is not the default behaviour. 

The repository does have documentation on why the `gen_crate` option doesn't work with buf remotely and what options do work.

https://github.com/neoeinstein/protoc-gen-prost/tree/main/protoc-gen-prost-crate#usage-with-buf

pkwarren:
@cobbinma - Is this plugin functional on its own without specifying any options? We set `no_include` on prost-serde and tonic just for the purposes of our testing in this repo (which runs each plugin individually), however the plugins are functional with the default options.

plugins/community/neoeinstein-prost-crate/v0.3.1/Dockerfile
```diff
@@ -0,0 +1,12 @@
+# syntax=docker/dockerfile:1.4
+FROM rust:1.68.0-alpine3.17 as builder
```
pkwarren, line 2:
```suggestion
FROM rust:1.70.0-alpine3.18 as builder
```

When introducing new plugin versions, we usually take it as an opportunity to bump to the latest Docker images.

tests/plugins_test.go
```diff
@@ -55,6 +55,7 @@ exec docker run --log-driver=none --rm -i {{.ImageName}}:{{.Version}} "$@"
 	// these plugins until the tests are updated to support running all plugin dependencies in sequence.
 	testOverrideOptions = map[string][]string{
 		"buf.build/community/neoeinstein-prost-serde": {"no_include"},
+		"buf.build/community/neoeinstein-prost-crate": {"no_features"},
```
pkwarren, line 58: nit: Can you move this up one line to keep these plugin names sorted?

cobbinma:
> @cobbinma - Is this plugin functional on its own without specifying any options? We set `no_include` on prost-serde and tonic just for the purposes of our testing in this repo (which runs each plugin individually), however the plugins are functional with the default options.

@pkwarren 
According to the usage with buf [documentation](https://github.com/neoeinstein/protoc-gen-prost/tree/main/protoc-gen-prost-crate#usage-with-buf) it should be, but it returns an error for me when running without options.

```sh
❯ buf generate --include-imports
WARN    Plugin "buf.build/prost/crate" is deprecated
Failure: plugin buf.build/prost/plugins/crate:v0.3.1-1: Cargo.toml: does not exist
```

pkwarren (approved):

maoueh:
Really cool, is it usable right now or there is some wait on Buf side?

pkwarren:
> Really cool, is it usable right now or there is some wait on Buf side?

It is available now and ready for use: https://buf.build/community/neoeinstein-prost-crate

There are limitations as documented here: https://github.com/neoeinstein/protoc-gen-prost/blob/main/protoc-gen-prost-crate/README.md. In particular, remote plugins don't have access to files from the filesystem, so the `gen_crate` option isn't supported.

maoueh:
Yep I'm aware thanks, I'm using it in the way it works.

PR #610, opened 2023-06-21
Add v1.115.0 stephenh-ts-proto plugin

cobbinma:
We still depend on `v1.115.0` of `stephenh-ts-proto`

Could this version be added so that we can move away from remote generation alpha

We are currently using this remote version:
```yaml
- remote: buf.build/stephenh/plugins/ts-proto:v1.115.0-0
    out: gen
```


mfridman:
I think this sounds reasonable. 

We intentionally set up this repository in a way to allow backfilling older plugin versions (on a case-by-case basis).

Everything looks good here, but need to commit the sum file. Every plugin in this repository is tested and verified against a few Buf modules.

If you have Docker locally, can you run:

```
make test PLUGINS="community/stephenh-ts-proto:v1.115.0"
```

plugins/community/stephenh-ts-proto/v1.115.0/package-lock.json
```diff
@@ -0,0 +1,327 @@
...
+        "": {
+            "name": "plugins-stephenh-ts-proto",
+            "version": "1.0.0",
+            "dependencies": {
+                "ts-proto": "1.115.0"
+            }
+        },
+        "node_modules/@protobufjs/aspromise": {
+            "version": "1.1.2",
+            "resolved": "https://perkbox-491933125842.d.codeartifact.eu-west-1.amazonaws.com/npm/frontend/@protobufjs/aspromise/-/aspromise-1.1.2.tgz",
```
mfridman, line 16: This isn't something we want to depend on.

plugins/community/stephenh-ts-proto/v1.115.0/package-lock.json
```diff
@@ -0,0 +1,327 @@
+{
+    "name": "plugins-stephenh-ts-proto",
+    "version": "1.0.0",
+    "lockfileVersion": 2,
```
mfridman, line 4: I don't think we are consistent with the `lockfileVersion`, and it's a mix of 2 and 3.

@pkwarren maybe we should consider adding a script to generate the lock files by invoking a docker container, that way we can pin the npm version and it'll be consistent. We can then verify the lock file was generated in a consistent way via CI. Just a thought (not blocking)

mfridman (approved):
lgtm, thanks for submitting a PR 🎉

plugins/community/stephenh-ts-proto/v1.115.0/package-lock.json
pkwarren: The fetch-versions script does this automatically. I do think we could re-run and warn if local diffs are detected.

PR #612, opened 2023-06-21
Add community/stephenh-ts-proto v1.115.0 

mfridman:


mfridman:
Added in https://github.com/bufbuild/plugins/pull/610

PR #622, opened 2023-06-28
Update to latest debian stable release

pkwarren:
Debian 12 (bookworm) is now stable. We should update our builds to start using this going forward.

### DIFF
--- a/.github/docker/Dockerfile.debian
+++ b/.github/docker/Dockerfile.debian
@@ -1,3 +1,3 @@
-FROM debian:bullseye-20230612
+FROM debian:bookworm-20230612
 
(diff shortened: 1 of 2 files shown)


.github/docker/Dockerfile.debian
```diff
@@ -1,3 +1,3 @@
-FROM debian:bullseye-20230612
+FROM debian:bookworm-20230612
```
mfridman, line 1: Hehe, I was looking at this the other day, and was waiting to see what the LTS schedule is.

https://wiki.debian.org/LTS

But, I trust you know better this world. And if you say it's stable, then so be it.

mfridman (approved):

PR #624, opened 2023-06-29
Add kotlin-stdlib in addition to jdk8 variant v2

pkwarren:
We should ensure that we pull in both kotlin-stdlib AND kotlin-stdlib-jdk8 to ensure that consistent versions are used with the dependencies and the compiler. Without this, it is possible that protobuf-kotlin's dependency version kotlin-stdlib 1.6.0 is used instead.

Breaking this change into multiple PRs as these plugins take a long time to build.


pkwarren:
Part 2 of 623.

stefanvanburen (approved):

PR #640, opened 2023-07-07
Update prost and tonic rust plugins

pkwarren:


mfridman (approved):

plugins/community/neoeinstein-prost/v0.2.3/buf.plugin.yaml
```diff
@@ -0,0 +1,9 @@
+version: v1
+name: buf.build/community/neoeinstein-prost
+plugin_version: v0.2.3
+source_url: https://github.com/neoeinstein/protoc-gen-prost
+description: Generates code using the Prost! code generation engine.
+output_languages:
+  - rust
+spdx_license_id: Apache-2.0
+license_url: https://github.com/neoeinstein/protoc-gen-prost/blob/1a6d3593622af18b75a4a79f545f8530cdaf444f/LICENSE
```
pkwarren, line 9: Noticed the link to the tag wasn't working so using this for now - opened https://github.com/neoeinstein/protoc-gen-prost/issues/74 upstream to see if we should use these going forward.

plugins/community/neoeinstein-prost/v0.2.3/Dockerfile
```diff
@@ -0,0 +1,12 @@
+# syntax=docker/dockerfile:1.4
+FROM rust:1.70.0-alpine3.18 as builder
+RUN apk add --no-cache musl-dev
+WORKDIR /app
+ENV CARGO_REGISTRIES_CRATES_IO_PROTOCOL=sparse
+RUN --mount=type=cache,target=/usr/local/cargo/registry,sharing=locked --mount=type=cache,target=/root/target \
```
pkwarren, line 6:
```suggestion
RUN --mount=type=cache,target=/usr/local/cargo/registry,sharing=locked --mount=type=cache,target=/root/target \
```

Added `sharing=locked` as apparently it isn't safe to have multiple writers to the Cargo registry (maybe a new regression in newer versions of Rust?).

mfridman: I guess this makes sense, at first I thought we could `sharing=private` so that each build creates a cached directory, but we want them to be re-used, just not in parallel?

mfridman (approved):

plugins/community/neoeinstein-prost/v0.2.3/Dockerfile
pkwarren: Favored this way to avoid multiple downloads of the same content, with the thought that they'd likely pull in a lot of the same dependencies.

PR #656, opened 2023-07-12
Add roadrunner-server php-grpc plugin

### LINKED ISSUE #655: Plugin request for Buf Schema Registry: RoadRunner `protoc-gen-php-grpc` plugin
## Mandatory

- Link to the repository: [link](https://github.com/roadrunner-server/grpc/tree/master/protoc_plugins)
- Previously used `GitHub` action to push the plugin: [link](https://github.com/roadrunner-server/roadrunner/blob/v2.12.0/.github/workflows/release_grpc_buf.yml)

## Optional

- `v2023.2.0`, version can be obtained from here: [link](https://github.com/roadrunner-server/roadrunner/releases)

- No

- No

pkwarren:
Fixes #655.


plugins/community/roadrunner-server-php-grpc/v4.2.0/Dockerfile
```diff
@@ -0,0 +1,12 @@
+# syntax=docker/dockerfile:1.4
+FROM golang:1.20.6-bullseye AS build
+RUN --mount=type=cache,target=/go/pkg/mod \
+    git clone --depth=1 --branch=v4.2.0 https://github.com/roadrunner-server/grpc.git \
```
pkwarren, line 4: Can't use `go install` here because the module has a `v2` w/ a v4.2.0 tag.

plugins/community/roadrunner-server-php-grpc/v4.2.0/Dockerfile
```diff
@@ -0,0 +1,12 @@
...
+    && cd grpc/protoc_plugins/protoc-gen-php-grpc \
+    && CGO_ENABLED=0 go build -ldflags "-s -w -X github.com/roadrunner-server/roadrunner/v2/internal/meta.version=v4.2.0" -trimpath
```
pkwarren, line 6: I don't know if this linker symbol is required, but I saw it in the Dockerfile that used to build the buf plugin.

plugins/community/roadrunner-server-php-grpc/v4.2.0/Dockerfile
rustatian: Yeah, I'll update the module version, sure.

pkwarren: Thanks @rustatian - not a huge deal but that would bring this more in line with other Go based protoc plugins we build here. I mainly left it as a note for our team.

plugins/community/roadrunner-server-php-grpc/v4.2.0/Dockerfile
rustatian: `gRPC` protoc plugin does not have its own version. Previously it was versioned according to the RoadRunner versions (like `v2023.1.0).
`-s -w` should be used separately. `-s` or `-w`, because `-s` remove symbol table with debug symbols, while `-w` removes only debug symbols, so their purpose overlaps. I prefer using `-s` for the release builds.

rustatian: But, I'm fine to tag plugin with the `gRPC` plugin (roadrunner-server/grpc) version, like `v4.2.0`, thank you for the super fast update 👍🏻

pkwarren: @rustatian - So should I remove the version symbol? I didn't see it used in the generated code - I'm guessing it is for a usage statement and not needed for a remote plugin?

pkwarren: Confirmed that it is used in the CLI / server in roadrunner and not referenced by the plugin, so I removed the custom linker symbol.

plugins/community/roadrunner-server-php-grpc/v4.2.0/Dockerfile
rustatian: Done: https://github.com/roadrunner-server/grpc/releases/tag/v4.3.0

plugins/community/roadrunner-server-php-grpc/v4.2.0/Dockerfile
rustatian: Yeah, right, it is used in the main RR repository, safe to delete, thanks 👍🏻

plugins/community/roadrunner-server-php-grpc/v4.2.0/Dockerfile
pkwarren: To use go install - i.e.:

```
$ go install -ldflags "-s -w" -trimpath github.com/roadrunner-server/grpc/protoc_plugins/v4/protoc-gen-php-grpc@v4.3.0
```

a module needs to have its own tag - i.e. `protoc_plugins/v4.3.0` - see https://github.com/grpc/grpc-go/tree/cmd/protoc-gen-go-grpc/v1.3.0/cmd/protoc-gen-go-grpc as an example. Alternatively, the CLI would need to move to the same module as the root of the project.

pkwarren: See https://go.dev/doc/modules/managing-source#multiple-module-source for more info.

rustatian: Oh, yeah, got u, forget about `go.work`, gimme a second

rustatian: Done, now should work 😃

plugins/community/roadrunner-server-php-grpc/v4.2.0/buf.plugin.yaml
```diff
@@ -0,0 +1,11 @@
+version: v1
+name: buf.build/community/roadrunner-server-php-grpc
+plugin_version: v4.2.0
```
mfridman, line 3: Are we going to use `v4.3.0` and update to `go install` ?

pkwarren: Updated.

mfridman (approved):
great job @pkwarren, some folks are going to be very happy (assuming the latest and greatest version is good enough).

PR #678, opened 2023-07-24
Update connect-kotlin and grpc-kotlin deps

pkwarren:
Update plugin dependencies for connect-kotlin and grpc-kotlin to pull in the latest version of the protocolbuffers/kotlin v23.4 plugin (which contains a patch for Kotlin package names).

### DIFF
--- a/plugins/bufbuild/connect-kotlin/v0.1.8/buf.plugin.yaml
+++ b/plugins/bufbuild/connect-kotlin/v0.1.8/buf.plugin.yaml
@@ -4,6 +4,7 @@ plugin_version: v0.1.8
 source_url: https://github.com/bufbuild/connect-kotlin
 description: Idiomatic gRPC & Connect RPCs for Kotlin.
 deps:
+  # v23.4 revision updated 20230724
   - plugin: buf.build/protocolbuffers/kotlin:v23.4
 output_languages:
   - kotlin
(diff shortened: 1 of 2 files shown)


plugins/bufbuild/connect-kotlin/v0.1.8/buf.plugin.yaml
```diff
@@ -4,6 +4,7 @@ plugin_version: v0.1.8
 source_url: https://github.com/bufbuild/connect-kotlin
 description: Idiomatic gRPC & Connect RPCs for Kotlin.
 deps:
+  # v23.4 revision updated 20230724
```
mfridman, line 7: Can you remind me, why we need this line?

mfridman: Ah right, to bump this so it depends on the newer patched revision. nvm

mfridman (approved):

plugins/bufbuild/connect-kotlin/v0.1.8/buf.plugin.yaml
pkwarren: This is needed to trigger a re-push of the plugin (with different buf.plugin.yaml), which when installed will automatically pull in the latest revision. Right now the previous version is installed and pins to the previous revision of the dependency.

Issue #694, opened 2023-08-01
Plugin request for Buf Schema Registry: cludden/protoc-gen-go-temporal


srs:
## Mandatory

**Where is the source code for the plugin?**

https://github.com/cludden/protoc-gen-go-temporal

## Optional

**Does the plugin have a valid semver version?**

Currently  `v1.0.0-beta.3`. Got it from https://github.com/cludden/protoc-gen-go-temporal/releases.

**Does the plugin have a dependency on another plugin?**

No.

**Does the output of this plugin depend on any other external libraries?**

Yes, it has runtime dependencies to the following Go modules:

* go.temporal.io/sdk/client
* go.temporal.io/sdk/workflow

PR #695, opened 2023-08-01
Added cludden/protoc-gen-go-temporal

srs:
I do not now if you update plugins manually or automatically. Anyway, to ease the task to add cludden/protoc-gen-go-temporal plugin I added a PR.

It's a great plugin that helps a lot with temporal workflows. Would appreaciate if you can add it to the community plugins.

This solves #694.

### DIFF
--- a/plugins/community/cludden-temporal-go/source.yaml
+++ b/plugins/community/cludden-temporal-go/source.yaml
@@ -1,0 +1,5 @@
+source:
+  github:
+    owner: cludden
+    repository: protoc-gen-go-temporal
+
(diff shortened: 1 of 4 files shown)


mfridman:
Once a PR is merged the added plugin(s) go into an automated workflow, and eventually get pushed to https://buf.build/plugins.

However, we're a bit selective with the public plugins we push to the BSR.

At this moment, it doesn't appear this is a widely-used plugin and we're unlikely to accept the PR, but we should keep the issue open and reevaluate if demand increases.

Side note, users on a paid plan can upload any Protobuf plugin and use them privately within their organization.
https://buf.build/docs/bsr/remote-plugins/custom-plugins

umutbasal (approved):

pkwarren:
Thank you for the contribution, but at the present we're going to close this PR. There is not a significant demand at the moment for this plugin. We may consider adding it at some point in the future.

In the meantime, users on paid plans could upload this as a custom plugin. See https://buf.build/docs/bsr/remote-plugins/custom-plugins for more details.

PR #747, opened 2023-08-16
Add connectrpc/es

smallsamantha:


smallsamantha:
opened this as an example, closing because @smaye81 is going to handle it

PR #750, opened 2023-08-16
Add new `connectrpc/query-es` plugin

smaye81:
This adds the @connectrpc/protoc-gen-connect-query plugin to the plugins repo.

Note this can't be merged until the package is officially published on npm.


smaye81:
@timostamm I named this `connectrpc/query-es` because I feel like `connectrpc/query` is a bit too vague and has a potential for name clashes down the road.

timostamm (commented):
> `connectrpc/query-es`

Agree with the name, considering other names are `go` and `es`.

timostamm:
Once packages are available in the `@connectrpc` scope, the lock file needs to be generated.

plugins/connectrpc/query-es/v0.4.2/package.json
```diff
@@ -0,0 +1,7 @@
+{
+  "name": "plugins-connectrpc-query-es",
+  "version": "1.0.0",
+  "dependencies": {
+    "@connectrpc/protoc-gen-connect-query": "0.4.2"
+  }
```
pkwarren, line 6: We should also check in a package-lock.json file alongside this.

pkwarren (approved):

PR #757, opened 2023-08-21
Update danielgtaylor/python-betterproto description

stefanvanburen:
Noticed this was the same as community/nipunn1313-mypy's description; I think it must have been missed when initially merging. Looked not right on https://buf.build/plugins.

No strong feelings on what this is changed to; for now, this is just taken from the top of the project's [README](https://github.com/danielgtaylor/python-betterproto#readme).


pkwarren (approved):
Doh - thank you!

mfridman (approved):

PR #770, opened 2023-08-28
Update workflow to only sync last day updates

pkwarren:
The total number of plugins are over 500+ (and the size to download them all is significant). Update the release workflow to only download plugins created/modified in the last 24h instead of all plugins since the beginning of time. This should greatly speed up the upload workflow for releases.

### DIFF
--- a/.github/workflows/upload.yml
+++ b/.github/workflows/upload.yml
@@ -22,14 +22,6 @@ jobs:
         with:
           go-version: '1.21'
           check-latest: true
-      - name: Cache Plugins
-        id: cache-plugins
-        uses: actions/cache@v3
-        with:
-          path: downloads
-          key: ${{ runner.os }}-plugin-downloads-${{ hashFiles('**/plugin-releases.json') }}
-          restore-keys: |
-            ${{ runner.os }}-plugin-downloads
         # uses https://cloud.google.com/iam/docs/workload-identity-federation to
         # swap a GitHub OIDC token for GCP service account credentials, allowing
         # this workflow to manage the buf-plugins bucket
@@ -44,7 +36,7 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ github.token }}
         run: |
-          go run ./cmd/download-plugins downloads
+          go run ./cmd/download-plugins -since 24h downloads
       - name: Upload To Release Bucket
         run: gsutil rsync -r downloads gs://buf-plugins
       - uses: dblock/create-a-github-issue@866beb009af3db457e82ca98efe474969a5ebce8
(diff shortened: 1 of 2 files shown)


.github/workflows/upload.yml
```diff
@@ -22,14 +22,6 @@ jobs:
...
           go-version: '1.21'
           check-latest: true
-      - name: Cache Plugins
-        id: cache-plugins
-        uses: actions/cache@v3
-        with:
-          path: downloads
-          key: ${{ runner.os }}-plugin-downloads-${{ hashFiles('**/plugin-releases.json') }}
-          restore-keys: |
-            ${{ runner.os }}-plugin-downloads
```
pkwarren, line 32: The cache got too large to be of any value - we're downloading the files from GitHub releases and so these should run quickly from the GH runners.

cmd/download-plugins/main.go
```diff
@@ -33,16 +34,24 @@ func run() error {
...
 
 	if len(flag.Args()) != 1 {
 		_, _ = os.Stderr.WriteString("usage: download-plugins <dir>\n")
 		flag.PrintDefaults()
 		os.Exit(2)
 	}
+
+	releaseSince := ""
+	if since > 0 {
+		releaseSince = fmt.Sprintf("%s.1", time.Now().UTC().Add(-since).Format("20060102"))
```
pkwarren, line 52: Include all releases starting with the first release of the day (`1`).

cmd/download-plugins/main.go
```diff
@@ -108,6 +117,9 @@ func run() error {
 				continue
 			}
 		}
+		if releaseSince != "" && pluginRelease.ReleaseTag < releaseSince {
```
pkwarren, line 120: Can use string comparison here as release tags are in the format `yyyyMMdd.<num>`.

stefanvanburen (approved):

pgmitche:
IIUC if the downloads directory is no longer cached, and is ephemeral now, if for some reason the upload action fails, we have the potential to completely miss pushing up some released versions until we notice/resolve the failure. At which point I suppose we'd have to remove the `-since` for one execution, which would download _everything_ and likely fail, or determine the time of the last successful run and adjust `-since {hours since last successful execution}` for one run?

pkwarren:
> IIUC if the downloads directory is no longer cached, and is ephemeral now, if for some reason the upload action fails, we have the potential to completely miss pushing up some released versions until we notice/resolve the failure. At which point I suppose we'd have to remove the `-since` for one execution, which would download _everything_ and likely fail, or determine the time of the last successful run and adjust `-since {hours since last successful execution}` for one run?

Yeah just the latter. A GH issue gets opened on every failure of the job so we should get notified right away. It is likely just tweaking from 24h -> 72h or something like that.

If we want more than 1 days worth of recovery, we can also just tweak the script to copy the last several days instead of the last 24h. Also note it runs on a schedule 4 times a day (every 6h).

PR #771, opened 2023-08-28
Make since parameter configurable

pkwarren:
In case we need to upload plugins further back than the default of 24h, add a workflow input to allow customizing the time.


pgmitche (approved):
LGreatTM 👍🏼

PR #772, opened 2023-08-29
Optimize plugin release time

pkwarren:
Swap out a CLI call to the docker CLI with a call to go-containerregistry API to fetch the same thing. This fixes a ~2s delay from checking for version information from every plugin (500+).


internal/cmd/release/main.go
```diff
@@ -582,38 +584,22 @@ func fetchRegistryImageAndImageID(plugin *plugin.Plugin) (string, string, error)
...
-	descriptorDigest := result.Descriptor.Digest
-	if descriptorDigest == "" {
-		return "", "", errors.New("unable to parse descriptor digest from docker manifest inspect output")
-	}
-	imageDigest := result.SchemaV2Manifest.Config.Digest
-	if imageDigest == "" {
-		return "", "", errors.New("unable to parse image config digest from docker manifest inspect output")
-	}
-	return fmt.Sprintf("%s@%s", imageName, descriptorDigest), imageDigest, nil
+	remoteImage, err := remote.Image(parsedName, remote.WithAuthFromKeychain(githubkeychain.Keychain))
```
pkwarren, line 592: This configures auth for ghcr.io container images with the GITHUB_TOKEN env var (we're already using this to authenticate with github).

internal/cmd/release/main.go
```diff
@@ -582,38 +584,22 @@ func fetchRegistryImageAndImageID(plugin *plugin.Plugin) (string, string, error)
...
+	}
+	manifest, err := remoteImage.Manifest()
+	if err != nil {
+		return "", "", err
+	}
+	remoteDigest, err := remoteImage.Digest()
+	if err != nil {
+		return "", "", err
+	}
+	return fmt.Sprintf("%s@%s", imageName, remoteDigest.String()), manifest.Config.Digest.String(), nil
```
pkwarren, line 604: Verified this returns the same information that the Docker CLI does:

```
...
2023/08/28 19:23:35 no changes to plugins since 20230826.1
```

rodaine (approved):

PR #775, opened 2023-08-30
Update connectrpc/es to latest bufbuild/es

pkwarren:


plugins/connectrpc/es/v0.13.2/buf.plugin.yaml
```diff
@@ -4,7 +4,7 @@ plugin_version: v0.13.2
 source_url: https://github.com/connectrpc/connect-es
 description: Generates code that is compatible with browsers and Node.js for working with Connect, gRPC, and gRPC-Web.
 deps:
-  - plugin: buf.build/bufbuild/es:v1.3.0
+  - plugin: buf.build/bufbuild/es:v1.3.1
```
pkwarren, line 7: These were released close to each other - before we publish this for the first time I thought we could update to the latest dependency.

stefanvanburen (approved):

PR #787, opened 2023-09-08
Configure buildx builder and enable local cache

pkwarren:
Update the Docker build steps to configure a buildx builder (so even when building locally we'll build with the docker-container driver just like the CI build). Update the dockerbuild command to use buildx local caching (and remove the need for keeping around each stage of a multi-stage build).


Makefile
```diff
@@ -29,7 +31,10 @@ all: build
 
 .PHONY: build
 build:
-	@go run ./internal/cmd/dockerbuild -org "$(DOCKER_ORG)" -- $(DOCKER_BUILD_EXTRA_ARGS)
+	docker buildx inspect "$(DOCKER_BUILDER)" 2> /dev/null || docker buildx create --use --bootstrap --name="$(DOCKER_BUILDER)"
```
pkwarren, line 34: Added this so local builds behave the same as ones in CI (they'll both use the `docker-container` driver).

internal/cmd/dockerbuild/main.go
```diff
@@ -64,18 +64,19 @@ func run(basedir string, dockerOrg string, args []string) error {
 			return err
 		}
 		var pluginKey string
+		const bazelPluginKey = "bazel"
```
pkwarren, line 67: This made a huge difference - having two bazel builds running at the same time caused major slowdowns. Here's an example after this change where you both see the caching and more reasonable build times: https://github.com/bufbuild/plugins/actions/runs/6117484022/job/16604171425

internal/docker/build.go
```diff
@@ -61,31 +67,6 @@ func Build(
 	defer func() {
 		retErr = errors.Join(retErr, f.Close())
 	}()
-	buildStages, err := ParseDockerfileBuildStages(f)
```
pkwarren, line 64: Now that we cache all the layers using buildx cache with `mode=max`, we don't need to do this any more.

mfridman (approved):
Thanks for digging 🏗️ into this.

PR #800, opened 2023-09-11
Create dockercache directory if it doesn't exist

pkwarren:


internal/docker/build.go
```diff
@@ -22,6 +22,10 @@ func Build(
 	dockerOrg string,
 	args []string,
 ) (_ []byte, retErr error) {
+	cacheDir := ".tmp/dockercache"
+	if err := os.MkdirAll(cacheDir, 0o755); err != nil {
```
mfridman, line 26:
```suggestion
	if err := os.MkdirAll(cacheDir, 0755); err != nil {
```

mfridman: Most of the std. lib and examples such as https://pkg.go.dev/os#FileMode use that notation, and not the octal one.

mfridman (approved):

PR #808, opened 2023-09-13
Update inputs for tibdex/github-app-token

pkwarren:
The previous 'repository' input has been replaced with two new inputs, 'installation_retrieval_mode' and 'installation_retrieval_payload'. The defaults for these values are the same as the previous 'repository' input, so we can remove this to fix warnings with the fetch versions workflow.

### DIFF
--- a/.github/workflows/fetch_versions.yml
+++ b/.github/workflows/fetch_versions.yml
@@ -22,7 +22,6 @@ jobs:
         with:
           app_id: 249762
           private_key: ${{ secrets.TOKEN_EXCHANGE_GH_APP_PRIVATE_KEY }}
-          repository: ${{ github.repository }}
           permissions: >-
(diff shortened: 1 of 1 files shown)


.github/workflows/fetch_versions.yml
```diff
@@ -22,7 +22,6 @@ jobs:
         with:
           app_id: 249762
           private_key: ${{ secrets.TOKEN_EXCHANGE_GH_APP_PRIVATE_KEY }}
-          repository: ${{ github.repository }}
```
pkwarren, line 25: https://github.com/tibdex/github-app-token/blob/95717386c657bc772ba69db6fafa0e08baf5372e/action.yml#L11-L30

mfridman (approved):

PR #812, opened 2023-09-14
Add grpc/* v1.58.1

mfridman:
This also makes a small change to the fetcher to be able to run the entire workflow locally. E.g.,

```bash
go run ./internal/cmd/fetcher/main.go plugins/grpc
```


internal/cmd/fetcher/main.go
```diff
@@ -128,11 +132,9 @@ func recreateNPMPackageLock(plugin createdPlugin) error {
 }
 
 // runPluginTests runs 'make test PLUGINS="org/name:v<new>"' in order to generate plugin.sum files.
-func runPluginTests(plugins []createdPlugin, rootDir string) error {
-	pluginsEnv := make([]string, 0, len(plugins)*2)
-	for _, plugin := range plugins {
-		pluginsEnv = append(pluginsEnv, fmt.Sprintf("%s/%s:%s", plugin.org, plugin.name, plugin.newVersion))
-	}
+func runPluginTests(plugin createdPlugin) error {
```
pkwarren, line 135: It would be better to have this execute tests on all plugins - otherwise we end up running common tests (unrelated to a specific plugin) potentially way more than necessary. What's the thinking behind making it run on an individual plugin?

internal/cmd/fetcher/main.go
```diff
@@ -144,14 +146,19 @@ func runPluginTests(plugins []createdPlugin, rootDir string) error {
...
+			fmt.Sprintf("PLUGINS=%s", pluginRef),
 		},
-		Dir:    rootDir,
-		Stdout: os.Stdout,
-		Stderr: os.Stderr,
-		Env:    env,
+		Dir: baseDir,
+		Env: env,
 	}
-	return cmd.Run()
```
mfridman, line 154: Maybe I'm misunderstanding how this works, but from trying this locally it masks the error, e.g.,

```bash
make: *** No rule to make target `test'.  Stop.
failed to run post-processing on plugins: exit status 2exit status 1
```

Whereas when you do `CombinedOutput()` and grab the returned string (even on an error) you get the underlying output, e.g.,

```
failed to run post-processing on plugins: failed to run plugin tests: exit status 2
docker buildx inspect "bufbuild-plugins" 2> /dev/null || docker buildx create --use --bootstrap --name="bufbuild-plugins"
ERROR: existing instance for "bufbuild-plugins" but no append mode, specify --node to make changes for existing instances
make: *** [build] Error 1
exit status 1
```

pkwarren (approved):

internal/cmd/fetcher/main.go
mfridman: I was under the impression we were running only tests required for testing / generating the hash file for a given plugin.

But this appears to be incorrect, we're calling `make test` which is go test./...

I don't get why we're running the entire test suite from within the fetcher, and not scoping it down to `go test ./tests `

mfridman: I'll change this to that slice though, as the point wrt to parallel testing is a good suggestion.

pkwarren (approved):

pkwarren (approved):

PR #814, opened 2023-09-14
Add workflow for arm64 builds

pkwarren:


.github/workflows/arm64.yml
```diff
@@ -0,0 +1,64 @@
...
+#      if: github.repository == 'bufbuild/plugins'
+#      uses: docker/login-action@465a07811f14bebb1938fbed4728c6a1ff8901fc
+#      with:
+#        registry: ghcr.io
+#        username: ${{ github.actor }}
+#        password: ${{ secrets.GITHUB_TOKEN }}
+    - name: Test
+      run: make test
+      env:
+        DOCKER_BUILD_EXTRA_ARGS: '--platform linux/arm64'
```
pkwarren, line 64: This should be enough to build everything for arm64. There may be some tweaks we can make (especially to Go builds) to make the builds a ton faster.

mfridman (approved):

PR #818, opened 2023-09-14
Add a per-plugin cache

pkwarren:
It appears than non-bazel builds aren't getting use of the local cache as well as other builds. Update the dockerbuild command to create unique cache directories per plugin and version, and reinstate the ability to build images in parallel.

### DIFF
--- a/internal/cmd/dockerbuild/main.go
+++ b/internal/cmd/dockerbuild/main.go
@@ -1,23 +1,39 @@
 package main
 
 import (
+	"cmp"
 	"context"
 	"errors"
 	"flag"
 	"fmt"
 	"log"
 	"os"
+	"path/filepath"
+	"runtime"
+	"slices"
 	"strings"
 	"time"
 
 	"github.com/bufbuild/buf/private/pkg/interrupt"
-
 	"github.com/bufbuild/plugins/internal/docker"
+	"golang.org/x/mod/semver"
+	"golang.org/x/sync/errgroup"
+
 	"github.com/bufbuild/plugins/internal/plugin"
 )
 
-// dockerbuild is a helper program used to build plugins from Dockerfiles.
+// dockerbuild is a helper program used to build plugins from Dockerfiles in an optimized fashion.
+// It replaces some clunky (and non-parallel) jobs in the Makefile.
+// It knows about relationships between common containers (like protoc/grpc Bazel plugins),
+// which enables optimized builds which build common code first before binaries.
 // It also makes it easier to add new labels to images using existing code to parse buf.plugin.yaml.
+
+const (
+	// larger amount of parallelism lead to OOM errors in testing - clamp for now.
+	maxLimit = 8
+	// plugin group for all bazel builds (they are run serially on their own as they are very resource intensive).
+	bazelPluginGroup = "bazel"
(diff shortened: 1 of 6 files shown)


internal/cmd/dockerbuild/main.go
```diff
@@ -1,55 +1,166 @@
...
 	if len(includedPlugins) == 0 {
 		return nil // nothing to build
 	}
-	for _, pluginToBuild := range includedPlugins {
+	pluginGroups, err := getPluginGroups(includedPlugins)
+	if err != nil {
+		return err
+	}
+	// Build bazel plugins on their own as they are very resource intensive.
+	if bazelPlugins, ok := pluginGroups[bazelPluginGroup]; ok {
```
pkwarren, line 87: Shuffled things around a bit so we build any bazel plugins first (serially). Then we build all the remaining ones in parallel up to the parallelism limit.

internal/cmd/dockerbuild/main.go
```diff
@@ -1,55 +1,166 @@
...
+		pluginGroup := pluginGroup
+		plugins := plugins
+		eg.Go(func() error {
+			return c.buildPluginGroup(ctx, pluginGroup, plugins)
+		})
+	}
+	return eg.Wait()
+}
+
+func getPluginGroups(plugins []*plugin.Plugin) (map[string][]*plugin.Plugin, error) {
```
pkwarren, line 121: Split this out into a helper method as it was getting unwieldy.

internal/cmd/dockerbuild/main.go
```diff
@@ -1,55 +1,166 @@
...
+		pluginGroups[pluginKey] = append(pluginGroups[pluginKey], pluginToBuild)
+	}
+	return pluginGroups, nil
+}
+
+func (c *command) buildPluginGroup(ctx context.Context, pluginGroup string, plugins []*plugin.Plugin) error {
+	for _, pluginToBuild := range plugins {
+		pluginIdentity := pluginToBuild.Identity
+		var pluginCacheDir string
+		// If cache is enabled, create a unique cache directory per group of plugins.
```
pkwarren, line 152: This is the new logic - we will create a unique cache directory for each plugin group. Here's an example:

```
.tmp/dockercache
├── bazel
│   ├── grpc
│   │   └── v1.58.1
│   └── protocolbuffers
│       └── v24.3
├── connectrpc
│   └── go
│       └── v1.11.1
└── grpc
    └── java
        └── v1.58.0
```

internal/cmd/release/main.go
```diff
@@ -182,10 +181,6 @@ func (c *command) calculateNewReleasePlugins(currentRelease *release.PluginRelea
 	var existingPlugins []release.PluginRelease
 
 	if err := plugin.Walk(c.rootDir, func(plugin *plugin.Plugin) error {
-		identity, err := bufpluginref.PluginIdentityForString(plugin.Name)
```
pkwarren, line 185: There were several places where we converted a plugin name to an identity. I updated this to happen when we load plugins so we don't have to do it (and handle errors) everywhere.

mfridman (approved):
This is looking great.

PR #827, opened 2023-09-19
Update to use slices package

pkwarren:


internal/plugin/plugin.go
```diff
@@ -85,12 +86,11 @@ func Walk(dir string, f func(plugin *Plugin) error) error {
 	}); err != nil {
 		return err
 	}
-	sort.SliceStable(unsorted, func(i, j int) bool {
-		p1, p2 := unsorted[i], unsorted[j]
-		if p1.Name != p2.Name {
-			return p1.Name < p2.Name
+	slices.SortStableFunc(unsorted, func(a, b *Plugin) int {
```
mfridman, line 89: Can you remind me, why SortStableFunc over SortFunc?

mfridman: In internal/cmd/release/main.go on line 280 we use `slices.SortFunc`, should we use the same sort function?

mfridman (approved):

internal/plugin/plugin.go
pkwarren: I don't think it matters at all here - I was just preserving the current logic which calls `SliceStable`.

mfridman (approved):

PR #844, opened 2023-10-03
Add protoc-gen-openapi v0.7.0

### LINKED ISSUE #821: Plugin request for Buf Schema Registry [protoc-gen-openapi]
## Mandatory

Example, the source code for the `protoc-gen-connect-go` plugin is found here:

https://github.com/google/gnostic/tree/main/cmd/protoc-gen-openapi

## Optional

Yes

cmd/protoc-json-openapi/v0.6.5

https://github.com/google/gnostic/releases/tag/cmd%2Fprotoc-json-openapi%2Fv0.6.5

Not to my knowledge.

Not to my knowledge.

@timburks please correct if necessary

mfridman:
Fix #821

### DIFF
--- a/plugins/community/google-gnostic-openapi/v0.7.0/Dockerfile
+++ b/plugins/community/google-gnostic-openapi/v0.7.0/Dockerfile
@@ -1,5 +1,5 @@
 # syntax=docker/dockerfile:1.6
-FROM golang:1.21.1-bullseye AS build
+FROM golang:1.21.2-bullseye AS build
 RUN --mount=type=cache,target=/go/pkg/mod \
     CGO_ENABLED=0 \
     go install -ldflags="-s -w" -trimpath github.com/google/gnostic/cmd/protoc-gen-openapi@v0.7.0
--- a/plugins/community/google-gnostic-openapi/v0.7.0/Dockerfile
+++ b/plugins/community/google-gnostic-openapi/v0.7.0/Dockerfile
@@ -1,0 +1,11 @@
+# syntax=docker/dockerfile:1.6
+FROM golang:1.21.1-bullseye AS build
+RUN --mount=type=cache,target=/go/pkg/mod \
+    CGO_ENABLED=0 \
+    go install -ldflags="-s -w" -trimpath github.com/google/gnostic/cmd/protoc-gen-openapi@v0.7.0
+
+FROM scratch
+COPY --from=build --link --chown=root:root /etc/passwd /etc/passwd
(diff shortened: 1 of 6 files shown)


plugins/community/google-gnostic-openapi/v0.7.0/buf.plugin.yaml
```diff
@@ -0,0 +1,9 @@
+version: v1
+name: buf.build/community/google-gnostic-openapi
```
mfridman, line 2:
```suggestion
name: buf.build/community/google-openapi
```

This alternative drops the "gnostic" from the name, which would be more in line with our existing nomenclature conventions of: `{owner}-{name}`

Where the repository is the upstream GitHub owner and the name is usually the name of the plugin minus `protoc-gen-` prefix.

However, for this particular one having the name "gnostic" in the name makes the most sense given the project's background. For some context:

- https://github.com/grpc-ecosystem/grpc-gateway/issues/441
- https://github.com/google/gnostic/issues/340

Although this is parked under github.com/gnostic organization,

mfridman:
If we have some cycles, we should consider adding a sample image that imports:

```
import "google/api/annotations.proto";
import "openapiv3/annotations.proto";
```

That way we exercise bits of the plugin that actually use the annotations.

plugins/community/google-gnostic-openapi/v0.7.0/Dockerfile
```diff
@@ -0,0 +1,11 @@
+# syntax=docker/dockerfile:1.6
+FROM golang:1.21.1-bullseye AS build
```
pkwarren, line 2:
```suggestion
FROM golang:1.21.2-bullseye AS build
```

plugins/community/google-gnostic-openapi/v0.7.0/buf.plugin.yaml
pkwarren: I prefer having the `gnostic` in the name too.

pkwarren:
> If we have some cycles, we should consider adding a sample image that imports:
> 
> ```
> import "google/api/annotations.proto";
> import "openapiv3/annotations.proto";
> ```
> 
> That way we exercise bits of the plugin that actually use the annotations.

We did something similar to verify the grpc-gateway code. I think it would be good to do the same here but at least there is an openapi.yaml file generated for our existing inputs (even if it is mostly empty).

pkwarren (approved):

PR #846, opened 2023-10-05
Add protocolbuffers/* v24.4

mfridman:


plugins/protocolbuffers/kotlin/v24.4/escape-package-names.patch
```diff
@@ -0,0 +1,64 @@
...
+     if (IsMapEntry(descriptor_->nested_type(i))) continue;
+diff --git a/src/google/protobuf/compiler/java/name_resolver.cc b/src/google/protobuf/compiler/java/name_resolver.cc
+index 0017e172a..b027865b1 100644
+--- a/src/google/protobuf/compiler/java/name_resolver.cc
++++ b/src/google/protobuf/compiler/java/name_resolver.cc
+@@ -358,6 +358,19 @@ std::string ClassNameResolver::GetKotlinExtensionsClassName(
+                           descriptor->file(), true, true, true);
+ }
+ 
++std::string ClassNameResolver::GetKotlinExtensionsClassNameEscaped(
```
mfridman, line 37: I can't remember why we have this patch. I wish we kept a list (or commented) why these exist and what specific problem they solve. And if there's any upstream issue we're tracking.

emcfarlane (approved):

plugins/protocolbuffers/java/v24.4/buf.plugin.yaml
```diff
@@ -0,0 +1,19 @@
...
+license_url: https://github.com/protocolbuffers/protobuf/blob/v24.4/LICENSE
+registry:
+  maven:
+    deps:
+      - com.google.protobuf:protobuf-java:3.24.4
+    additional_runtimes:
+      - name: lite
+        deps:
+          - com.google.protobuf:protobuf-javalite:3.24.4
+          - build.buf:protobuf-javalite:3.24.4
```
mfridman, line 18: I forgot we have a dependency on this. And we should avoid publishing java and co. before the javalite version is available which is in a separate repository

mfridman: This has been released (thank you), 😌 

https://repo1.maven.org/maven2/build/buf/protobuf-javalite/

PR #880, opened 2023-10-27
Update node runtime image to match build image

pkwarren:


plugins/bufbuild/es/v1.4.0/Dockerfile
```diff
@@ -1,11 +1,11 @@
 # syntax=docker/dockerfile:1.4
-FROM node:20.8.1-alpine3.18 AS build
+FROM node:20.9.0-alpine3.18 AS build
```
pkwarren, line 2: Really only cosmetic changes - felt safe to pull it in: https://github.com/nodejs/node/compare/v20.8.1...v20.9.0

PR #892, opened 2023-11-02
Add protoc v25.0 plugins

pkwarren:


pkwarren:
Verified generated code diffs from v24.4 - you can see the results here: https://gist.github.com/pkwarren/57f08a2437e2d07ffe2eb1bd76686fef

stefanvanburen (approved):
nice, excellent that we can remove that patch now that it's upstreamed!

PR #899, opened 2023-11-06
Add initial Python Repository config

stefanvanburen:
Companion to https://github.com/bufbuild/buf/pull/2544. We can hold off on merging this until we feel good about the structure, and the buf PR lands in main.

For fetching updates of the buf definitions, can use:

```
$ env GOPROXY=direct go get -u github.com/bufbuild/buf@svanburen/python-repository-config
```


plugins/protocolbuffers/pyi/v24.4/buf.plugin.yaml
```diff
@@ -7,3 +7,11 @@ output_languages:
   - python
 spdx_license_id: BSD-3-Clause
 license_url: https://github.com/protocolbuffers/protobuf/blob/v24.4/LICENSE
+registry:
```
stefanvanburen, line 10: note to myself to move pb/pyi and pb/python to the v25.0 version (or whatever the latest one is when we want to ultimately merge this)

go.mod
```diff
@@ -4,7 +4,7 @@ go 1.21
 
 require (
 	aead.dev/minisign v0.2.0
-	github.com/bufbuild/buf v1.27.2
+	github.com/bufbuild/buf v1.27.3-0.20231106140608-d7a87496b879
```
pkwarren, line 7: We'll also need to update the .github/workflow files for CI to build with the newer buf version (@mfridman just had to do this for another new field added to buf.plugin.yaml).

plugins/community/nipunn1313-mypy-grpc/v3.5.0/buf.plugin.yaml
```diff
@@ -7,3 +7,22 @@ spdx_license_id: Apache-2.0
...
+      - "types-protobuf >= 4.23.0.2"
+      - "protobuf >= 4.23.4"
+      # https://github.com/nipunn1313/mypy-protobuf/tree/v3.5.0#to-run-typecheckers-on-code-generated-with-grpc-plugin---youll-additionally-need
+      # > Earlier releases might work, but aren't tested
+      # TODO: Should we just remove the version specifiers, then?
+      - "grpcio >= 1.56.2"
+      # https://pypi.org/project/grpcio-tools/
+      - "grpcio-tools >= 1.56.2"
+      # https://pypi.org/project/grpc-stubs/
+      - "grpc-stubs >= 1.53.0.2"
```
pkwarren, line 28: Over time what is the best way to keep these up to date with newer dependencies? Is it something we can have the fetcher do as it upgrades things?

stefanvanburen: Unfortunately, there's not a great way that I'm aware of - there's not a way that these are consistently encoded across Python packages. In this _particular_ package's case, we might be able to read them from the [`test_requirements.txt`](https://github.com/nipunn1313/mypy-protobuf/blob/main/test_requirements.txt#L6-L8) file, but other plugins (like the protocolbuffer/{python,pyi} ones) don't have any encoded information about dependencies to use the generated code.

One other wrinkle with Python dependencies is that libraries are typically expected to keep their dependencies unpinned - that is, not use `==`. They can set a minimum version (like what we're doing here), but this _does_ mean that we're now forcing users using this package to install a version greater than or equal to what we're specifying, which isn't necessarily a strict limit.

That's one of the reasons why I have the `TODO` above here - it's not clear to me if we should even include the `>=`, or leave it to consumers to specify the version of these deps that they want?

go.mod
stefanvanburen: makes sense - looks like the revert was this PR: https://github.com/bufbuild/plugins/pull/886, so when the downstream PR is merged in buf I can update these back to pulling from `main` (assuming we don't get out a release in the meantime)

plugins/protocolbuffers/pyi/v24.4/buf.plugin.yaml
stefanvanburen: Added to v25.0 versions in https://github.com/bufbuild/plugins/pull/899/commits/85f860a49c719ee821f18988072abc2d4d175c4b

go.mod
stefanvanburen: This should be released in https://github.com/bufbuild/buf/releases/tag/v1.28.0 - I'm assuming we'll get that pulled in by Dependabot next week?

plugins/community/nipunn1313-mypy-grpc/v3.5.0/buf.plugin.yaml
stefanvanburen: I've been thinking through this off-and-on this week and really haven't come up with much better. My preference would be that we start with:

* use the ranges specified by the plugin author, if available
  * we can relax these if people complain and want to figure out their own versions, see below
* otherwise, leave the version range open and leave it to the user to install the version they need.
  * we can tighten these up if we realize that there's a plugin that _always_ requires something from a certain version of a dependency

Obviously, we can/should tighten these up if we get more information, but ISTM that this is just sorta the convention in the ecosystem: libraries should declare pretty loose requirements so that upstream applications can figure out the versions they need. (See https://hynek.me/articles/semver-will-not-save-you/ section "Unintended Consequence: Version Conflicts" for a bit of related reading, but the core thing I'm gesturing at is leaving it up to the user to figure these things out.)

plugins/protocolbuffers/pyi/v24.4/buf.plugin.yaml
```diff
@@ -7,3 +7,12 @@ output_languages:
...
+  python:
+    package_type: "stub-only"
+    # https://github.com/protocolbuffers/protobuf/blob/v24.4/python/setup.py#L471
+    requires_python: ">=3.7"
+    deps:
+      # https://pypi.org/project/protobuf/
+      - "protobuf"
+      - "types-protobuf"
```
pkwarren, line 18: What do you think about just making these changes to start on the latest versions of the plugins?

pkwarren (approved):
Looks good.

plugins/protocolbuffers/pyi/v24.4/buf.plugin.yaml
stefanvanburen: reverted in https://github.com/bufbuild/plugins/pull/899/commits/ef3fb6865918d718e4f8bd72b67f08e0abaa2d1e

pkwarren (approved):

PR #906, opened 2023-11-10
Only rebuild plugins if changed

pkwarren:
Update the dockerbuild command to add a label containing the git revision of the plugin's directory (if there are no changes to the working directory of the plugin). If we determine that an image exists for the same revision of the plugin, skip costly rebuilds.

Additionally, update commands to use exec.CommandContext consistently and stop calling exec.LookPath as the exec package does this automatically for us if the command doesn't contain path separators.

### DIFF
--- a/internal/cmd/fetcher/main.go
+++ b/internal/cmd/fetcher/main.go
@@ -15,6 +15,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/bufbuild/buf/private/pkg/interrupt"
 	"go.uber.org/multierr"
 	"golang.org/x/mod/semver"
 
@@ -34,12 +35,13 @@ func main() {
 		os.Exit(2)
 	}
 	root := os.Args[1]
-	created, err := run(context.Background(), root)
+	ctx, _ := interrupt.WithCancel(context.Background())
+	created, err := run(ctx, root)
 	if err != nil {
 		_, _ = fmt.Fprintf(os.Stderr, "failed to fetch versions: %v\n", err)
 		os.Exit(1)
 	}
-	if err := postProcessCreatedPlugins(created); err != nil {
+	if err := postProcessCreatedPlugins(ctx, created); err != nil {
 		_, _ = fmt.Fprintf(os.Stderr, "failed to run post-processing on plugins: %v\n", err)
 		os.Exit(1)
 	}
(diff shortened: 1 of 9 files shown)


internal/docker/build.go
```diff
@@ -40,12 +36,19 @@ func Build(
...
 		fmt.Sprintf("org.opencontainers.image.description=%s", plugin.Description),
 		"--label",
 		fmt.Sprintf("org.opencontainers.image.licenses=%s", plugin.SPDXLicenseID),
+		"--label",
+		fmt.Sprintf("org.opencontainers.image.vendor=%s", "Buf Technologies, Inc."),
 		"--progress",
 		"plain",
 	}
+	if gitCommit := plugin.GitCommit(ctx); gitCommit != "" {
+		commonArgs = append(commonArgs, "--label", fmt.Sprintf("org.opencontainers.image.revision=%s", gitCommit))
```
pkwarren, line 50: If there aren't uncommitted changes to the plugin's directory, we'll add the full git revision as an image label.

internal/plugin/plugin.go
```diff
@@ -291,6 +297,48 @@ func (p IncludePlugin) Matches(pluginName, pluginVersion, latestVersion string)
 	return p.version == pluginVersion
 }
 
+// GitCommit calculates the last git commit for the plugin's directory.
+// This will return an empty string if there are uncommitted changes to the plugin's directory.
+// This is used to label the built Docker image and also avoid unnecessary Docker builds.
+func (p *Plugin) GitCommit(ctx context.Context) string {
+	p.gitCommitOnce.Do(func() {
+		if gitModified, err := calculateGitModified(ctx, p.Path); err != nil {
+			log.Printf("failed to calculate git modified status: %v", err)
```
pkwarren, line 306: At some point I'd like to convert all this over to `slog` but for now this will do.

internal/cmd/release/main.go
```diff
@@ -504,30 +502,16 @@ func createPluginReleases(dir string, plugins []release.PluginRelease) error {
...
-		Args: append([]string{
-			dockerPath,
-			command,
-		}, args...),
-		Stdout: os.Stdout,
-		Stderr: os.Stderr,
-	}
-	return cmd, nil
+func dockerCmd(ctx context.Context, command string, args ...string) *exec.Cmd {
+	cmd := exec.CommandContext(ctx, "docker", append([]string{command}, args...)...) //nolint:gosec
```
stefanvanburen, line 511: I'm assuming all these gosec lints are just due to possibly `exec`ing user-given args, which we're not concerned about?

internal/plugin/plugin.go
```diff
@@ -28,10 +31,13 @@ type Plugin struct {
...
 	// Plugin identity (parsed from ExternalConfig.Name).
 	Identity bufpluginref.PluginIdentity `yaml:"-"`
+	// For callers that need git commit info - ensure we only calculate it once.
+	gitCommitOnce sync.Once `yaml:"-"`
+	gitCommit     string    `yaml:"-"`
 }
 
 func (p *Plugin) String() string {
-	return fmt.Sprintf("%+v", *p)
+	return fmt.Sprintf("%s:%s", p.Identity.IdentityString(), p.Version)
```
stefanvanburen, line 40: I'm assuming nothing was relying on the old `String` output? (Just for debugging?)

stefanvanburen (approved):

internal/cmd/fetcher/main.go
```diff
@@ -85,24 +88,17 @@ func runGoModTidy(plugin createdPlugin) error {
 		// no go.mod/go.sum to update
 		return nil
 	}
-	goPath, err := exec.LookPath("go")
```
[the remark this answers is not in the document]
stefanvanburen: 🎉 these `cmd` cleanups are nice, + the context integration

internal/cmd/release/main.go
pkwarren: Yeah exactly. These are internal commands used by the build process.

internal/plugin/plugin.go
pkwarren: Yep was just for debugging. It broke when I added a sync.Once here.

PR #930, opened 2023-11-29
Start tracking distroless image updates

pkwarren:
In order to improve reproducibility of builds, we should start to track distroless image updates. Add docker files for each flavor of distroless images we're using (additional follow up work is necessary to the fetch versions command to replace old versions with new ones).

Update debian/golang to bookworm going forward and update swift to use a 22.04 base image instead of the 20.04 one.

### DIFF
--- a/.github/docker/Dockerfile.debian
+++ b/.github/docker/Dockerfile.debian
@@ -1,3 +1,3 @@
-FROM debian:bullseye-20231120
+FROM debian:bookworm-20231120
 
 CMD echo this is a dummy file used to automate dependency upgrades for plugins
--- a/.github/docker/Dockerfile.distroless-base-debian
+++ b/.github/docker/Dockerfile.distroless-base-debian
@@ -1,0 +1,3 @@
+FROM gcr.io/distroless/base-debian12:latest@sha256:1dfdb5ed7d9a66dcfc90135b25a46c25a85cf719b619b40c249a2445b9d055f5
+
+CMD echo this is a dummy file used to automate dependency upgrades for plugins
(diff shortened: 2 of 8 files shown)


.github/docker/Dockerfile.debian
```diff
@@ -1,3 +1,3 @@
-FROM debian:bullseye-20231120
+FROM debian:bookworm-20231120
```
pkwarren, line 1: I know we tried this before and some things failed but now there have been a few minor releases and I'd expect newer versions to be compatible.

.github/docker/Dockerfile.distroless-base-debian
```diff
@@ -0,0 +1,3 @@
+FROM gcr.io/distroless/base-debian12:latest@sha256:1dfdb5ed7d9a66dcfc90135b25a46c25a85cf719b619b40c249a2445b9d055f5
```
pkwarren, line 1: Distroless images don't use tags so hoping that dependabot will be able to update these revisions.

.github/docker/Dockerfile.distroless-java-debian
```diff
@@ -0,0 +1,3 @@
+FROM gcr.io/distroless/java17-debian12:latest@sha256:8f80873debfafc0b77dd22d03e6d34d0a716c52404ca4ca19f76f2de11000c55
```
pkwarren, line 1: We currently have java11 and java17 in use - I think for the minimal places we run jar files we can safely upgrade to a newer JVM.

mfridman (approved):

PR #978, opened 2023-12-15
Update docker dependabot updates to weekly

mfridman:


.github/dependabot.yml
```diff
@@ -12,4 +12,4 @@ updates:
   - package-ecosystem: "docker"
     directory: "/.github/docker"
     schedule:
-      interval: "daily"
+      interval: "weekly"
```
pkwarren, line 15: I haven't minded the frequency so far, except for this week which had two batches of distroless updates (after several weeks with no changes).

What do you think about maybe grouping the distroless ones together but still keeping daily frequency so we are pretty up to date on any security issues?

mfridman: Sure, that'd work for me.

PR #1022, opened 2024-01-24
Fix Dockerfile for community/mercari-grpc-federation

goccy:
Hi @mfridman , Thank you for adding the plugin the other day 😄 

I tried to use our plugin and found that the `protoc-gen-grpc-federation` command depends on the `go` command. This is because we use the [imports](https://pkg.go.dev/golang.org/x/tools/imports) package to format generated code.

So, I fixed Dockerfile to allow the use of the `go` command. Please check it.

The following is the error output.

```
: err: go command required, not found: exec: "go": executable file not found in $PATH: stderr:
```

The following error occurs if `GOROOT` ENV is not set, so that setting is also added.

```
: err: exit status 2: stderr: go: cannot find GOROOT directory: 'go' binary is trimmed and GOROOT is not set
```


mfridman:
Ouch, I didn't realize this plugin invoked the Go command. 

Note, when using [protogen](https://pkg.go.dev/google.golang.org/protobuf/compiler/protogen) it handles formatting with the [go/printer](https://pkg.go.dev/go/printer) package. Specifically here:

https://github.com/protocolbuffers/protobuf-go/blob/b70f02be537926b5693bb047189608a7f2c80994/compiler/protogen/protogen.go#L1126-L1129

Having said all that, I'd prefer the plugin to be updated so it didn't need any dependencies. We enforce a strict contract for the plugins we maintain, i.e., they must operate entirely with the [CodeGeneratorRequest](https://buf.build/protocolbuffers/wellknowntypes/docs/main:google.protobuf.compiler#google.protobuf.compiler.CodeGeneratorRequest) / [CodeGeneratorResponse](https://buf.build/protocolbuffers/wellknowntypes/docs/main:google.protobuf.compiler#google.protobuf.compiler.CodeGeneratorResponse) and not need any external tools or files to function.

Sorry about that, I should have caught this earlier.

tests/testdata/buf.build/community/mercari-grpc-federation/v0.9.2/grpc-federation/plugin.sum
```diff
@@ -0,0 +1 @@
+h1:d0P6WfW6ve00/zKIclE7fO8vZGCrQGOZnkhw9papifI=
```
mfridman, line 1: This is the sum of the generated code from the sample image

```go
// Code generated by protoc-gen-grpc-federation. DO NOT EDIT!
package gen

import (
	"context"
	"fmt"
	"io"
	"log/slog"
	"runtime/debug"

	"github.com/google/cel-go/cel"
	celtypes "github.com/google/cel-go/common/types"
	grpcfed "github.com/mercari/grpc-federation/grpc/federation"
	grpcfedcel "github.com/mercari/grpc-federation/grpc/federation/cel"
	"go.opentelemetry.io/otel"
	"go.opentelemetry.io/otel/trace"
	grpccodes "google.golang.org/grpc/codes"
	grpcstatus "google.golang.org/grpc/status"
)

// Federation_GetPostResponseArgument is argument for "federation.GetPostResponse" message.
type Federation_GetPostResponseArgument[T any] struct {
	Id     string
	Client T
}

// FederationServiceConfig configuration required to initialize the service that use GRPC Federation.
type FederationServiceConfig struct {
	// Resolver provides an interface to directly implement message resolver and field resolver not defined in Protocol Buffers.
	// If this interface is not provided, an error is returned during initialization.
	Resolver FederationServiceResolver // required
	// ErrorHandler Federation Service often needs to convert errors received from downstream services.
	// If an error occurs during method execution in the Federation Service, this error handler is called and the returned error is treated as a final error.
	ErrorHandler grpcfed.ErrorHandler
	// Logger sets the logger used to output Debug/Info/Error information.
	Logger *slog.Logger
}
[...]
```

pkwarren (approved):

goccy:
@mfridman OK, Thank you for your reviewing. I will fix this in the next update. BTW, I knew about those formatters, but they did not match our use case 😅  . ( They cannot optimize the list of import packages )

mfridman:
For my curiosity, what do you mean by _"optimize the list of import packages"_?

But thanks for your understanding, once this is resolved please let us know so we can flip this config and resume pulling the latest versions, (it is currently disabled):

https://github.com/bufbuild/plugins/blob/e22244eb441fdb0a4be04ce4d93380778200e8a5/plugins/community/mercari-grpc-federation/source.yaml#L5

goccy:
> For my curiosity, what do you mean by "optimize the list of import packages"?

The packages used in the results generated by the gRPC Federation vary depending on the context. Therefore, a list of necessary and sufficient packages are predefined, and unnecessary packages are removed by the formatter.

PR #1017, opened 2024-01-19
Add mercari/grpc-federation plugin v0.9.2

mfridman:
Fix #991

PR #1024, opened 2024-01-24
Update test case for empty dir hash in gen dir

mfridman:
This PR adds a defensive check in our plugin tests to avoid passing plugins that generate no code and produce an "empty dir" hash.

There is a historical allowlist of plugins we'll allow, but moving forward we should craft protos to exercise valid code generation, in a similar way as:

https://github.com/bufbuild/plugins/blob/d524f5ffe0e922ac10752db571b60a320f7fc7f4/tests/plugins_test.go#L133-L13

For context, this would help catch https://github.com/bufbuild/plugins/pull/1017 against the typical images we pass in testing.

> --- FAIL: TestGeneration (0.42s)
    --- FAIL: TestGeneration/plugins/community/mercari-grpc-federation/v0.9.2 (0.00s)
        --- FAIL: TestGeneration/plugins/community/mercari-grpc-federation/v0.9.2/eliza (0.83s)
            plugins_test.go:99: generated code directory is empty for buf.build/community/mercari-grpc-federation:v0.9.2
        --- FAIL: TestGeneration/plugins/community/mercari-grpc-federation/v0.9.2/petapis (0.84s)
            plugins_test.go:99: generated code directory is empty for buf.build/community/mercari-grpc-federation:v0.9.2
FAIL
FAIL    github.com/bufbuild/plugins/tests       1.852s
FAIL
make: *** [test] Error 1

### DIFF
--- a/internal/plugin/plugin.go
+++ b/internal/plugin/plugin.go
@@ -37,7 +37,7 @@ type Plugin struct {
 }
 
 func (p *Plugin) String() string {
-	return fmt.Sprintf("%s:%s", p.Identity.IdentityString(), p.Version)
+	return fmt.Sprintf("%s:%s", p.Identity.IdentityString(), p.PluginVersion)
 }
 
 // Dependency represents a dependency one plugin has on another.
--- a/tests/plugins_test.go
+++ b/tests/plugins_test.go
@@ -83,6 +83,12 @@ func TestGeneration(t *testing.T) {
 			bufCmd.Dir = pluginDir
 			output, err := bufCmd.CombinedOutput()
 			require.NoErrorf(t, err, "buf generate failed - output: %s", string(output))
+			// Ensure the gen directory is not empty, otherwise we'll get a sum of an empty directory.
+			genDirFiles, err := os.ReadDir(pluginGenDir)
+			require.NoError(t, err, "failed to read generated code directory")
+			if len(genDirFiles) == 0 {
+				t.Fatal("generated code directory is empty")
+			}
 			genDirHash, err := dirhash.HashDir(pluginGenDir, "", dirhash.Hash1)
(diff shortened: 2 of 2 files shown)


internal/plugin/plugin.go
```diff
@@ -37,7 +37,7 @@ type Plugin struct {
 }
 
 func (p *Plugin) String() string {
-	return fmt.Sprintf("%s:%s", p.Identity.IdentityString(), p.Version)
+	return fmt.Sprintf("%s:%s", p.Identity.IdentityString(), p.PluginVersion)
```
mfridman, line 40: Version here is `v1` of the config, it should be the plugin version.

pkwarren: Doh - good catch!

tests/plugins_test.go
```diff
@@ -58,6 +58,12 @@ exec docker run --log-driver=none --rm -i {{.ImageName}}:{{.Version}} "$@"
 		"buf.build/community/neoeinstein-prost-serde": {"no_include"},
 		"buf.build/community/neoeinstein-tonic":       {"no_include"},
 	}
+	// Some plugins do not generate any code for the test protos, so we allow an empty plugin.sum file for these
+	// plugins.
+	allowedEmptyPluginSums = map[string]bool{
+		"buf.build/bufbuild/validate-java": true,
+		"buf.build/grpc-ecosystem/gateway": true,
```
pkwarren, line 65: Do we want to add an exception for the`buf.build/community/mercari-grpc-federation` plugin for now until a fix is made available (at least for the current version)?

mfridman: I want to undo and remove that plugin entirely, and re-open the issue advocating for an upstream fix. Any concerns with that?

PR #1029, opened 2024-01-26
Add community/mfridman-go-json v1.2.0

mfridman:
This PR adds a new community plugin which supersedes the existing one mitchellh/protoc-gen-go-json. I've disabled further updates of that plugin, and will mark it deprecated shortly.

This should also resolve the upstream issue folks have been asking about. https://github.com/mitchellh/protoc-gen-go-json/issues/17

### DIFF
--- a/plugins/community/mfridman-go-json/source.yaml
+++ b/plugins/community/mfridman-go-json/source.yaml
@@ -1,0 +1,4 @@
+source:
+  github:
+    owner: mfridman
+    repository: protoc-gen-go-json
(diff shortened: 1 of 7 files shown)


plugins/community/mfridman-go-json/source.yaml
```diff
@@ -0,0 +1,4 @@
+source:
+  github:
+    owner: mfridman
```
pkwarren, line 3: Hmm... not sure about this maintainer. 😀

pkwarren (approved):

plugins/community/mfridman-go-json/source.yaml
mfridman: 😰

PR #1116, opened 2024-03-14
Add protoc v26.0 and protobuf-ts v2.9.4

pkwarren:


.github/docker/Dockerfile.bazel
```diff
@@ -1,3 +1,3 @@
-FROM gcr.io/bazel-public/bazel:7.1.0
+FROM gcr.io/bazel-public/bazel:7.0.2
```
pkwarren, line 1: Something the new bazel is doing is breaking protoc builds. I tested some locally and couldn't find a solution so reverting this back for now.

Build failures look like:
```
$ bazel build '//plugins:protoc-gen-cpp.stripped'
ERROR: Failed to initialize sandbox: /home/build/.cache/bazel/_bazel_build/433461e849f916304f45cf88ae617024/sandbox/_moved_trash_dir -> /home/build/.cache/bazel/_bazel_build/433461e849f916304f45cf88ae617024/sandbox/stale-trash-0 (Invalid cross-device link)
INFO: Elapsed time: 0.198s
INFO: 0 processes.
ERROR: Build did NOT complete successfully
```

plugins/protocolbuffers/java/v26.0/buf.plugin.yaml
```diff
@@ -0,0 +1,19 @@
...
+source_url: https://github.com/protocolbuffers/protobuf
+description: Base types for Java. Generates message and enum types.
+output_languages:
+  - java
+spdx_license_id: BSD-3-Clause
+license_url: https://github.com/protocolbuffers/protobuf/blob/v26.0/LICENSE
+registry:
+  maven:
+    deps:
+      - com.google.protobuf:protobuf-java:4.26.0
```
pkwarren, line 13: Major version of dependencies bumped for Java.

plugins/protocolbuffers/kotlin/v26.0/buf.plugin.yaml
```diff
@@ -0,0 +1,27 @@
...
+  - kotlin
+spdx_license_id: BSD-3-Clause
+license_url: https://github.com/protocolbuffers/protobuf/blob/v26.0/LICENSE
+registry:
+  maven:
+    compiler:
+      kotlin:
+        version: 1.8.22
+    deps:
+      - com.google.protobuf:protobuf-kotlin:4.26.0
```
pkwarren, line 19: Major version bump for kotlin too.

plugins/protocolbuffers/pyi/v26.0/buf.plugin.yaml
```diff
@@ -0,0 +1,18 @@
...
+spdx_license_id: BSD-3-Clause
+license_url: https://github.com/protocolbuffers/protobuf/blob/v26.0/LICENSE
+registry:
+  python:
+    package_type: "stub-only"
+    # https://github.com/protocolbuffers/protobuf/blob/v26.0/python/dist/setup.py#L78
+    requires_python: ">=3.8"
+    deps:
+      # https://pypi.org/project/protobuf/
+      - "protobuf"
```
pkwarren, line 17: @stefanvanburen - In light of https://protobuf.dev/support/cross-version-runtime-guarantee/#major, I wonder if we don't want to be more constrained here. The upstream package bumped from major version 4 to 5 with this release.

pkwarren (commented):
Initial commit contains the initial code from fetcher. https://github.com/bufbuild/plugins/pull/1116/commits/305a8665e2c459ad6097d2fdf83967d6509f9b47 are my deltas on top.

plugins/protocolbuffers/pyi/v26.0/buf.plugin.yaml
stefanvanburen: yeah, was just looking at this. I think you're right. Debating between a strict `==` specifier or using the `~=` compatible specifier ([docs](https://packaging.python.org/en/latest/specifications/version-specifiers/#id5)); thoughts based on what we use elsewhere?

stefanvanburen: i.e. I think we should either pin directly to `==5.26.0`, or do something like `~=5.26` (so `5.26.0` or `5.26.1` are valid, but not `5.27`) or `~=5` for `5.*`.

pkwarren: I think using the compatible specifier would be good to start. On the Java side we pull in an exact version but users can pull in a later version when consuming the artifacts.

stefanvanburen: yeah. Maybe `~=5` to match anything within the major version, since in theory that's where we'd get in trouble with incompatibilities?

pkwarren: Settled on `~= 5.26` for now which should be a best effort to pull in a compatible version of the runtime but not be too constrained.

mfridman (approved):

PR #1118, opened 2024-03-15
Move typescript module in a separate layer

pkwarren:
All of the @bufbuild/protoplugin based plugins include typescript for support for transpiling .ts to .js/.d.ts. For most plugins, the ~62M typescript module ends up being the majority of the entire plugin + node_modules directory (~64M).

If we optimize the docker builds for these plugins, we can split the typescript module off into its own layer. This will benefit pushing to OCI registries (since layers will be shared across images) and pulling images to execute images (less layers to fetch).

There may be additional optimations to take for these plugins but this shows pretty decent benefits locally (pushing all compressed layers to a local docker registry reduces the registry storage size from 97.5M to 65.3M). This will continue to pay benefits going forward as long as the underlying typescript module version isn't bumped.


plugins/bufbuild/es/v1.8.0/Dockerfile
```diff
@@ -1,11 +1,13 @@
-# syntax=docker/dockerfile:1.4
+# syntax=docker/dockerfile:1.7
 FROM node:20.11.1-alpine3.19 AS build
 WORKDIR /app
 COPY --link package*.json .
-RUN npm ci
+RUN npm ci \
```
stefanvanburen, line 5: tangentially related, but should we be running `npm ci` with `NODE_ENV=production` or an `--omit` to avoid installing devDependencies? It definitely won't get similar gains to this, but might trim `node_modules` even further, unless I'm misunderstanding. (not an expert on node, could be very wrong.)

stefanvanburen (approved):
slick

plugins/bufbuild/es/v1.8.0/Dockerfile
pkwarren: Yep - that would be a good idea. However, we set up package.json with a single dependency which is the NPM package for the plugin - for example: https://github.com/bufbuild/plugins/blob/main/plugins/bufbuild/es/v1.8.0/package.json.

pkwarren:
Moving this to a draft - continuing to do some investigation here.

pkwarren:
Updated to use esbuild to build plugin (excluding typescript dependency) which strips most plugins from ~3M -> 500K. Strip typescript dependency down to only imported file (62M -> 10M). Continue to add the typescript dependency in a separate layer for re-use with other plugins. This should essentially make fetching a new plugin just pulling a 500K layer when we update versions (assuming we haven't bumped base images or typescript).

Before:
```
bufbuild/plugins-connectrpc-query-es   v1.3.1            9cac72d82d79   7 minutes ago   247MB
bufbuild/plugins-bufbuild-es           v1.8.0            7884c162cf7a   7 minutes ago   248MB
bufbuild/plugins-connectrpc-es         v1.4.0            d7b7da1e8a91   7 minutes ago   247MB
bufbuild/plugins-bufbuild-knit-ts      v0.0.7            cc83fa7d2df6   7 minutes ago   247MB
```

After:
```
bufbuild/plugins-bufbuild-knit-ts      v0.0.7            706e9f8c3b2a   2 minutes ago    192MB
bufbuild/plugins-connectrpc-query-es   v1.3.1            f3a9e7356d85   3 minutes ago    192MB
bufbuild/plugins-bufbuild-es           v1.8.0            3e3b97f7ccf3   5 minutes ago    192MB
bufbuild/plugins-connectrpc-es         v1.4.0            7c64e74f690a   5 minutes ago    192MB
```

This also reduced the compressed .zip plugins (with all docker layers) from ~64M to ~52M. They don't today have the benefit of sharing layers with other plugins in the zip files, but will see speedups when pushing plugins to an OCI registry or pulling them.

plugins/bufbuild/es/v1.8.0/Dockerfile
pkwarren: Ended up taking this one step further and building the plugin with esbuild.

pkwarren:
Other optimizations appear possible for other node plugins:

* community/stephenh-ts-proto
  * Can be build with esbuild with `./node_modules/.bin/esbuild ./node_modules/.bin/protoc-gen-ts_proto --bundle --platform=node --external:dprint-node --outfile=protoc-gen-ts_proto.js`.
  * The `dprint-node` package ships rust binaries for all supported architectures (we should only need to keep the one for the current architecture). We could also ship it as a separate layer like we do for typescript:
    * Arch: darwin-arm64, darwin-x64, linux-arm64-gnu, linux-arm64-musl, linux-x64-gnu, linux-x64-musl, win32-x64-msvc
* community/timostamm-protobuf-ts
  * Can use same optimizations as the `@bufbuild/protoplugin` based plugins here.

pkwarren:
Included a workaround for https://github.com/GoogleContainerTools/distroless/issues/1548 and now image sizes are down even further:

```
REPOSITORY                                         TAG               IMAGE ID       CREATED          SIZE
bufbuild/plugins-bufbuild-knit-ts                  v0.0.7            42fe1b17b6cf   30 seconds ago   140MB
bufbuild/plugins-connectrpc-es                     v1.4.0            bfdde4dc162f   30 seconds ago   140MB
bufbuild/plugins-community-timostamm-protobuf-ts   v2.9.4            c7fdbbeeb326   30 seconds ago   139MB
bufbuild/plugins-bufbuild-es                       v1.8.0            7c9e792de01e   30 seconds ago   140MB
bufbuild/plugins-connectrpc-query-es               v1.3.1            2c5cb4845170   30 seconds ago   140MB
```

plugins/community/timostamm-protobuf-ts/v2.9.4/Dockerfile
```diff
@@ -1,11 +1,18 @@
...
+FROM gcr.io/distroless/nodejs20-debian12:latest@sha256:e0edb89411e3f93863e9bd97c15cddcbac968d7a16b7a82af6ae6b264da65de9 AS node
+
+FROM gcr.io/distroless/cc-debian12:latest@sha256:e6ae66a5a343d7112167f9117c4e630cfffcd80db44e44302759ec13ddd2d22b
+COPY --link --from=node --chmod=0755 /nodejs/bin/node /nodejs/bin/node
+COPY --link --from=build --chmod=0755 /app/protoc-gen-ts.js /app/protoc-gen-ts.js
+COPY --link --from=build /app/node_modules/typescript /app/node_modules/typescript
 USER nobody
-ENTRYPOINT [ "/app/node_modules/.bin/protoc-gen-ts" ]
+ENTRYPOINT ["/nodejs/bin/node"]
+CMD [ "/app/protoc-gen-ts.js" ]
```
pkwarren, line 18: Verified this plugin with the option `--output_javascript` as well to ensure that the transpile process continues to work.

mfridman (approved):
This feels a bit risky, and potentially a maintenance burden. But willing to give this a try.

stefanvanburen (approved):

PR #1122, opened 2024-03-15
Pass -m option to gsutil rsync

pkwarren:
Optimize rsync job to pass the -m option to fix a build warning shown in the release logs.

### DIFF
--- a/.github/workflows/upload.yml
+++ b/.github/workflows/upload.yml
@@ -50,7 +50,7 @@ jobs:
         run: |
           go run ./cmd/download-plugins -since ${{ inputs.since }} downloads
       - name: Upload To Release Bucket
-        run: gsutil rsync -r downloads gs://buf-plugins
+        run: gsutil -m rsync -r downloads gs://buf-plugins
       - uses: dblock/create-a-github-issue@866beb009af3db457e82ca98efe474969a5ebce8
         if: failure()
         env:


.github/workflows/upload.yml
```diff
@@ -50,7 +50,7 @@ jobs:
         run: |
           go run ./cmd/download-plugins -since ${{ inputs.since }} downloads
       - name: Upload To Release Bucket
-        run: gsutil rsync -r downloads gs://buf-plugins
+        run: gsutil -m rsync -r downloads gs://buf-plugins
```
pkwarren, line 53: Saw this printed in logs:
```
==> NOTE: You are performing a sequence of gsutil operations that may
run significantly faster if you instead use gsutil -m rsync ... Please
see the -m section under "gsutil help options" for further information
about when gsutil -m can be advantageous.
```

stefanvanburen: cool - looks like this is just parallelizing things? from `gsutil help options`:

```
  -m          Causes supported operations (acl ch, acl set, cp, mv, rm, rsync,
              and setmeta) to run in parallel. This can significantly improve
              performance if you are performing operations on a large number of
              files over a reasonably fast network connection.
```

stefanvanburen (approved):

PR #1150, opened 2024-04-03
Add googlecloudplatform/protoc-gen-bq-schema

mfridman:
Draft for now, as there is no upstream version to tag off of. But got all tests tests passing, and the resulting file it generates:

matches the upstream example.

https://github.com/GoogleCloudPlatform/protoc-gen-bq-schema/tree/31a7e43419f7c19d79de0bb506798bc602287e80?tab=readme-ov-file#example-with-policy-tags

tests/testdata/buf.build/googlecloudplatform/bq-schema/v0.0.1/bq-schema/gen

```json
[
 {
  "name": "a",
  "type": "INTEGER",
  "mode": "REQUIRED",
  "policyTags": {
   "names": [
    "private"
   ]
  }
 },
 {
  "name": "b",
  "type": "STRING",
  "mode": "NULLABLE",
  "policyTags": {
   "names": [
    "public"
   ]
  }
 },
 {
  "name": "nested",
  "type": "RECORD",
  "mode": "REQUIRED",
  "fields": [
   {
    "name": "a",
    "type": "INTEGER",
    "mode": "REQUIRED",
    "policyTags": {
     "names": [
      "private"
     ]
    }
   },
   {
    "name": "b",
    "type": "STRING",
    "mode": "NULLABLE"
   }
  ]
 }
]
```


plugins/googlecloudplatform/bq-schema/v1.1.0/Dockerfile
```diff
@@ -0,0 +1,15 @@
...
+
+RUN --mount=type=cache,target=/go/pkg/mod \
+    go install -ldflags="-s -w" -trimpath github.com/GoogleCloudPlatform/protoc-gen-bq-schema@v1.1.0 \
+    && mv /go/bin/${GOOS}_${GOARCH}/protoc-gen-bq-schema /go/bin/protoc-gen-bq-schema || true
+
+FROM scratch
+COPY --from=build --link --chown=root:root /etc/passwd /etc/passwd
+COPY --from=build --link /go/bin/protoc-gen-bq-schema /
+USER nobody
+ENTRYPOINT [ "/protoc-gen-bq-schema", "-logtostderr=true" ]
```
pkwarren, line 15: A bit surprised that logging to stderr is not the default (since otherwise it would emit a malformed generator response).

pkwarren (approved):

plugins/googlecloudplatform/bq-schema/v1.1.0/Dockerfile
```diff
@@ -0,0 +1,15 @@
+# syntax=docker/dockerfile:1.6
+FROM --platform=$BUILDPLATFORM golang:1.22.1-bookworm AS build
```
pkwarren, line 2:
```suggestion
FROM --platform=$BUILDPLATFORM golang:1.22.2-bookworm AS build
```

plugins/googlecloudplatform/bq-schema/v1.1.0/Dockerfile
mfridman: I'm not too familiar with [golang/glog](https://github.com/golang/glog), but if I removed this it would fail for me locally. Let me try to remove it and see if it fails in CI.

plugins/googlecloudplatform/bq-schema/v1.1.0/Dockerfile
mfridman: Updated in https://github.com/bufbuild/plugins/pull/1150/commits/e73e9c56052bf48907ff5e35afe551340ca38017

pkwarren (approved):

PR #1151, opened 2024-04-03
Add community/mercari-grpc-federation v0.13.6

mfridman:
Fix https://github.com/bufbuild/plugins/issues/1148


tests/plugins_test.go
```diff
@@ -150,7 +150,7 @@ func TestGeneration(t *testing.T) {
 					testPluginWithImage(t, toTest, "grpc-federation")
 				}
 				if semver.Compare(toTest.PluginVersion, "v0.11.0") >= 0 {
-					testPluginWithImage(t, toTest, "grpc-federation-v0.11.0")
+					testPluginWithImage(t, toTest, "grpc-federation-latest")
```
mfridman, line 153: On second thought, maybe we should just keep all existing up-to the newly built image and not worry about anything breaking.

mfridman: added explicit image for v0.13.6 and higher.

pkwarren (approved):

PR #1211, opened 2024-04-30
Update grpc/go plugin to latest security release

### LINKED ISSUE #1210: Update google.golang.org/grpc
Hi 👋 

Would you be able to update [google.golang.org/grpc](https://github.com/bufbuild/plugins/blob/main/plugins/grpc/go/v1.3.0/buf.plugin.yaml#L15) to the latest [release](https://github.com/grpc/grpc-go/releases)

Would be great to use a version that is not effected by the [gRPC-Go HTTP/2 Rapid Reset](https://github.com/advisories/GHSA-m425-mq94-257g) vulnerability

pkwarren:
Update grpc/go to require the earliest version of the google.golang.org/grpc dependency with security fixes (v1.61.2). Update the plugin dependency to the latest version of protocolbuffers/go. Improve dockerfile to support multi-arch builds.

Fixes #1210.

### DIFF
--- a/plugins/grpc/go/v1.3.0/Dockerfile
+++ b/plugins/grpc/go/v1.3.0/Dockerfile
@@ -1,5 +1,9 @@
-# syntax=docker/dockerfile:1.4
-FROM golang:1.20.1-bullseye AS build
+# syntax=docker/dockerfile:1.7
+FROM --platform=$BUILDPLATFORM golang:1.22.2-bookworm AS build
+
+ARG TARGETOS TARGETARCH
+ENV CGO_ENABLED=0 GOOS=$TARGETOS GOARCH=$TARGETARCH
+
 WORKDIR /tmp
 RUN git clone --depth 1 --branch cmd/protoc-gen-go-grpc/v1.3.0 https://github.com/grpc/grpc-go.git
 WORKDIR /tmp/grpc-go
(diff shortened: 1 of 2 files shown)


plugins/grpc/go/v1.3.0/Dockerfile
```diff
@@ -69,11 +73,10 @@ index 340eaf3e..7f0fba56 100644
 EOF
 WORKDIR /tmp/grpc-go/cmd/protoc-gen-go-grpc
 RUN --mount=type=cache,target=/go/pkg/mod \
-    CGO_ENABLED=0 \
-    go build -ldflags "-s -w" -trimpath
+    go build -o protoc-gen-go-grpc -ldflags "-s -w" -trimpath
```
mfridman, line 76: Should we update this to be like the other go plugins, i.e., `&& mv /go/bin/${GOOS}_${GOARCH}`

pkwarren: This one is different because we're patching it to support outputting to a separate package.

mfridman (approved):

plugins/grpc/go/v1.3.0/Dockerfile
mfridman: Hehe, as soon as I hit reply, I remembered about the patch 😅

PR #1212, opened 2024-04-30
Add dependency on prost for tonic

stefanvanburen:
Similar to grpc/go and connectrpc/go depending on protocolbuffers/go, tonic relies on prost codegen for the base message types.

Just adding this to the latest tonic version for now, but happy to go back and add it elsewhere (either here or as a follow-up).

### DIFF
--- a/plugins/community/neoeinstein-tonic/v0.4.0/buf.plugin.yaml
+++ b/plugins/community/neoeinstein-tonic/v0.4.0/buf.plugin.yaml
@@ -3,6 +3,8 @@ name: buf.build/community/neoeinstein-tonic
 plugin_version: v0.4.0
 source_url: https://crates.io/crates/protoc-gen-tonic
 description: Generates Tonic gRPC server and client code using the Prost! code generation engine.
+deps:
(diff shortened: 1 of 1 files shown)


plugins/community/neoeinstein-tonic/v0.4.0/buf.plugin.yaml
```diff
@@ -3,6 +3,8 @@ name: buf.build/community/neoeinstein-tonic
 plugin_version: v0.4.0
 source_url: https://crates.io/crates/protoc-gen-tonic
 description: Generates Tonic gRPC server and client code using the Prost! code generation engine.
+deps:
+  - plugin: buf.build/community/neoeinstein-prost:v0.3.1
```
stefanvanburen, line 7: https://buf.build/community/neoeinstein-prost

pkwarren: What about prost-serde? Does it have a dependency on prost too?

pkwarren (approved):

PR #1223, opened 2024-05-07
Update grpc-gateway patch

### LINKED ISSUE #1200: Investigate import issue with grpc-ecosystem/gateway plugin
Likely related to https://github.com/grpc-ecosystem/grpc-gateway/issues/3801

The Generated SDK code is generating wrong imports and https://github.com/grpc-ecosystem/grpc-gateway/issues/3801#issuecomment-2072703665 explains fairly well how to reproduce it. Based on that structure when the `separate_package=true` option is enabled it builds an incorrect import in `service1/v1/service/service1v1gateway, specifically

```diff
- "buf.build/gen/go/.../service2/v1/service2v1grpc"
+ "buf.build/gen/go/.../service1/v1/service1v1grpc"
```

I suspect it's using a filename instead of the full path, but that's just a hunch.

~### Debug notes:~

~✅ strategy: directory~
~❌ strategy: all~

mfridman:
Fix #1200

See https://github.com/bufbuild/plugins/issues/1200#issuecomment-2099284521 for more details.

- [x] Test this with BSR


plugins/grpc-ecosystem/gateway/v2.19.1/separate_pkg_additional_imports.patch
```diff
@@ -195,26 +183,30 @@ index 62766254..033d1c09 100644
...
  		}
 +		goPkg := file.GoPkg
 +		fileNamePrefix := file.GeneratedFilenamePrefix
 +		if g.separatePackage {
 +			goPkg = descriptor.GoPackage{
-+				Path: filepath.Join(file.GoPkg.Path, file.GoPkg.Name),
++				Path: path.Join(file.GoPkg.Path, file.GoPkg.Name),
 +				Name: file.GoPkg.Name,
 +			}
-+			fileNamePrefix = path.Join(file.GeneratedFilenamePrefix, file.GoPkg.Name, filepath.Base(file.GeneratedFilenamePrefix))
```
mfridman, line 217: This will unfortunately be a breaking change, but this appears like a bug in the previous implementation. We were including the filename as a directory and then concatenating with the grpcgateway directory:

### Before

```
service1
└── v1
    ├── service
    │   └── service1v1gateway
    │       └── service.pb.gw.go
    ├── service.pb.go
    └── service1v1grpc
        └── service_grpc.pb.go
```

### After

```
service1
└── v1
    ├── service.pb.go
    ├── service1v1gateway
    │   └── service.pb.gw.go
    └── service1v1grpc
        └── service_grpc.pb.go
```

plugins/grpc-ecosystem/gateway/v2.19.1/separate_pkg_additional_imports.patch
```diff
@@ -11,102 +72,30 @@ index 1cfc4b02..11e1368d 100644
...
-+						baseTypePackageSubPath,
-+						grpcPackageSubPath,
-+						1,
-+					),
-+					Name: strings.Replace(packageName, baseTypePackageSubPath, grpcPackageSubPath, 1),
-+				},
-+			}
-+			additionalImports = append(additionalImports, svc.GRPCFile.GoPkg.Path)
-+		}
-+		r.SetAdditionalImports(additionalImports)
```
mfridman, line 109: The bug reported in #1200 was related to this bit of code. We are looping over all files and all services, but only adding additional imports based on the current service (and overriding any previously set ones). 

So if you have multiple proto packages with multiple services, you'd end up with an incorrect imports when they are added later on.

plugins/grpc-ecosystem/gateway/v2.19.1/buf.plugin.yaml
```diff
@@ -16,7 +16,7 @@ registry:
     - standalone=true
     - separate_package=true
 deps:
-  - plugin: buf.build/protocolbuffers/go:v1.32.0
+  - plugin: buf.build/protocolbuffers/go:v1.34.0
```
pkwarren, line 19:
```suggestion
  - plugin: buf.build/protocolbuffers/go:v1.34.1
```

plugins/grpc-ecosystem/gateway/v2.19.1/separate_pkg_additional_imports.patch
pkwarren: Hopefully this won't be too much of an impact to people - the new package layout is definitely what we intended it to be originally and it matches how grpc and connect generate code.

pkwarren (approved):

plugins/grpc-ecosystem/gateway/v2.19.1/separate_pkg_additional_imports.patch
mfridman: I've thought long about this, and as much as I dislike breaking changes, I'd argue this is a bug fix at the expense of a quick import fix for end-users.

It'll also make the grpc-gateway implementation consistent with the grpc and connect plugins. And if we have additional plugins that generate to a separate directory they should all follow a convention, which leaves this grpc-gateway plugin as an outlier.

Lastly, explaining generated SDK paths is sometimes challenging to end-users. Having this non-conventional path is yet another barrier to adoption, at least for grpc-gateway.

cc @bufdev what do you think?

The alternative is to leave the generated imports as-is and remember to fix this in a major version only.

mfridman: @pkwarren reverted the breaking change in https://github.com/bufbuild/plugins/pull/1223/commits/2dba0c1b7fe1bafb999d7556f1ba073284a00a7a

For the gateway, we continue to build the same path as before. I left a comment to remind us of this in case we want to address this in the future.

pkwarren (approved):

PR #1231, opened 2024-05-10
Add initial Cargo Registry configuration

stefanvanburen:
This adds initial Cargo Registry support for `community/neoeinstein-prost/v0.3.1` and `community/neoeinstein-tonic/v0.4.0`. We can backport to previous versions if requested.


go.mod
```diff
@@ -4,7 +4,7 @@ go 1.22.0
 
 require (
 	aead.dev/minisign v0.2.1
-	github.com/bufbuild/buf v1.31.0
+	github.com/bufbuild/buf v1.31.1-0.20240508184409-79f9211ad6f0
```
stefanvanburen, line 7: pulling in https://github.com/bufbuild/buf/pull/2956 which hasn't been released yet.

plugins/community/neoeinstein-prost/v0.3.1/buf.plugin.yaml
```diff
@@ -7,3 +7,24 @@ output_languages:
...
 spdx_license_id: Apache-2.0
 license_url: https://github.com/neoeinstein/protoc-gen-prost/blob/protoc-gen-prost-v0.3.1/LICENSE
+registry:
+  cargo:
+    # https://github.com/neoeinstein/protoc-gen-prost/blob/protoc-gen-prost-v0.3.1/protoc-gen-prost/Cargo.toml#L21
+    # https://github.com/tokio-rs/prost/blob/v0.12.3/prost-build/Cargo.toml#L15
+    # TODO: Should we actually set this? This is just the rust-version set by
+    # prost-build, which isn't necessarily the rust-version that the *generated
+    # code* requires.
+    rust_version: "1.60"
```
stefanvanburen, line 17: Left a couple of comments in here for the sake of discussion in this PR that I'm going to rip out.

To start off with, for both of these plugins, it's generally pretty difficult to understand what the generated code needs, both as a MSRV and dependencies.

For the MSRV, I think we should leave it unset for both plugins, for now. Determining the MSRV of a regular Rust crate is already tricky. [Here's the section describing the field](https://doc.rust-lang.org/cargo/reference/manifest.html#the-rust-version-field) from the reference, which links to [this somewhat scary-looking suggestion of using third-party tools to verify that the specified `rust-version` is _actually_ the MSRV for the crate](https://doc.rust-lang.org/cargo/guide/continuous-integration.html#verifying-rust-version). Similar to Python's `python-version`, unless the plugin or underlying generator specifically documents and tests the MSRV for generated code, I doubt we should populate this value.

plugins/community/neoeinstein-prost/v0.3.1/buf.plugin.yaml
```diff
@@ -7,3 +7,24 @@ output_languages:
...
+    deps:
+      # https://github.com/tokio-rs/prost/blob/v0.12.3/prost-build/Cargo.toml#L30
+      # TODO: What features are required by the generated code?
+      - name: "prost"
+        req: "0.12.3"
+        default_features: false
+      # prost-types is necessary for any module using the WKTs.
+      - name: "prost-types"
+        req: "0.12.3"
+        default_features: false
```
stefanvanburen, line 27: `prost-build` will generate code that depends on the `prost` crate, and any code using the WKT (at least, with the default options set), will depend on `prost-types`.

Another option is [setting the `compile_well_known_types` option](https://github.com/neoeinstein/protoc-gen-prost/tree/protoc-gen-prost-v0.3.1/protoc-gen-prost#options), which would allow us to drop the `prost-types` dependency, but we initially decided on doing the former; I think we should stick with that for now.

The only other concern I have here is the version requirements: we're currently specifying `0.12.3`, which in [Rust-parlance actually means a range of versions](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-cratesio) — translating, this means this package supports `0.12.3  :=  >=0.12.3, <0.13.0`. I _guess_ that's fine, but given that the *plugin* is released separately from the underlying generator, I'm not sure if we'd run into situations where e.g. `prost@0.13.0` is released, and users want to upgrade to that version, but the plugin lags in release and thus users aren't able to upgrade.

(Again, I think this is fine for now — mostly leaving notes for the future.)

plugins/community/neoeinstein-tonic/v0.4.0/buf.plugin.yaml
```diff
@@ -9,3 +9,20 @@ output_languages:
...
+    # tonic-build@v0.11.0 does not specify a rust-version.
+    # TODO: Maybe remove the above comment(s), as it's not important what
+    # version tonic-build specifies, but what the generated code requires.
+    rust_version: ""
+    deps:
+      - name: "tonic"
+        req: "0.11.0"
+        # https://github.com/hyperium/tonic/blob/v0.11.0/tonic/Cargo.toml#L29
+        # TODO: What features are required by the generated code?
+        default_features: false
```
stefanvanburen, line 25: Similar comments to above; tricky to understand the deps / rust-version of the generated code, so intentionally simple for now. Going to ask about this on `tonic-build` repo.

plugins/community/neoeinstein-tonic/v0.4.0/buf.plugin.yaml
```diff
@@ -9,3 +9,20 @@ output_languages:
...
+  # https://github.com/neoeinstein/protoc-gen-prost/blob/protoc-gen-tonic-v0.4.0/protoc-gen-tonic/README.md#options
+  opts:
+    - no_include=true
```
stefanvanburen, line 28: We set this option so that the `tonic` plugin does not attempt to read files that it assumes to have been created by the `prost` plugin.

plugins/community/neoeinstein-prost/v0.3.1/buf.plugin.yaml
mfridman: Agreed, even more so, if it's optional we should just exclude it unless there's a good reason (like it breaks a build) to include it.

plugins/community/neoeinstein-tonic/v0.4.0/buf.plugin.yaml
mfridman: Yeah, reminds me of this goodie

https://github.com/bufbuild/plugins/blob/4a30a8e5e7842443efb518bf6f7533fbaae448df/tests/plugins_test.go#L53-L59

mfridman (approved):

PR #1244, opened 2024-05-16
Add stephenh-ts-proto v1.176.0

### LINKED ISSUE #1243: Updated release of stephenh/ts-proto?
I see that regular updates were disabled due to the frequency of release. However, newer versions of the plugin contain generator changes that would be nice to use via `buf`.

pkwarren:
Fixes #1243.


plugins/community/stephenh-ts-proto/v1.176.0/Dockerfile
```diff
@@ -0,0 +1,19 @@
+# syntax=docker/dockerfile:1.7
+FROM node:20.13.1-alpine3.19 AS build
+WORKDIR /app
+COPY --link package*.json .
+RUN npm ci
+RUN sed -i -e 's|/usr/bin/env node|/nodejs/bin/node|g' /app/node_modules/ts-proto/protoc-gen-ts_proto \
+ && rm -f /app/node_modules/dprint-node/*darwin*.node /app/node_modules/dprint-node/*win32*.node
```
pkwarren, line 7: Partial cleanup of https://github.com/bufbuild/plugins/pull/1118#issuecomment-2000279316. If we want to get more sophisticated we can remove all versions except the one we're building for (`linux-arm64-gnu` or `linux-x64-gnu`), however just starting with removing darwin/windows.

plugins/community/stephenh-ts-proto/v1.176.0/Dockerfile
```diff
@@ -0,0 +1,19 @@
...
+FROM gcr.io/distroless/nodejs20-debian12:latest@sha256:a69b487c1d156e3aeaab0ffb587f46248c8e891e289081a3c28f7104a69c4515 AS node
+
+FROM gcr.io/distroless/cc-debian12:latest@sha256:e6ae66a5a343d7112167f9117c4e630cfffcd80db44e44302759ec13ddd2d22b AS base
+
+FROM scratch
+COPY --link --from=base / /
+COPY --link --from=node --chmod=0755 /nodejs/bin/node /nodejs/bin/node
+COPY --link --from=build /app /app
+USER nobody
+ENTRYPOINT ["/nodejs/bin/node"]
```
mfridman, line 18: Can you remind me why we need 2 entrypoints here?

plugins/community/stephenh-ts-proto/v1.176.0/Dockerfile
```diff
@@ -0,0 +1,19 @@
...
+RUN npm ci
+RUN sed -i -e 's|/usr/bin/env node|/nodejs/bin/node|g' /app/node_modules/ts-proto/protoc-gen-ts_proto \
+ && rm -f /app/node_modules/dprint-node/*darwin*.node /app/node_modules/dprint-node/*win32*.node
+
+FROM gcr.io/distroless/nodejs20-debian12:latest@sha256:a69b487c1d156e3aeaab0ffb587f46248c8e891e289081a3c28f7104a69c4515 AS node
+
+FROM gcr.io/distroless/cc-debian12:latest@sha256:e6ae66a5a343d7112167f9117c4e630cfffcd80db44e44302759ec13ddd2d22b AS base
+
+FROM scratch
+COPY --link --from=base / /
```
pkwarren, line 14: Testing squashing on base image layers based on @DMarby example. This reduces total number of layers in the image from 18 -> 3 (base image, node binary, plugin).

plugins/community/stephenh-ts-proto/v1.176.0/Dockerfile
pkwarren: We're using a base image of cc, not node (which sets this entrypoint). This is so we can avoid shipping all the include files and other unused files from the node image.

mfridman (approved):
I'm in favor of this approach 👍, nice job!

PR #1266, opened 2024-05-29
Backport Python `protobuf` compatible versions to v25.*

stefanvanburen:
We're already doing this for the v26 ([e.g.](https://github.com/bufbuild/plugins/blob/313fe18533139a4b3a70516a3082de038c6c4a67/plugins/protocolbuffers/python/v26.1/buf.plugin.yaml#L17)) and v27 plugins. v25 is the earliest release currently supported by Generated SDKs, and is on the [4.x releases](https://pypi.org/project/protobuf/#history) of `protobuf` — this just makes it easier for users to pin to these versions when other dependencies rely on a lower version of `protobuf`.

v25 also happens to be the earliest version of the plugins with Generated SDK support, so now we're covering all of our existing bases with these compatible version specifiers.

### DIFF
--- a/plugins/protocolbuffers/python/v25.0/buf.plugin.yaml
+++ b/plugins/protocolbuffers/python/v25.0/buf.plugin.yaml
@@ -14,4 +14,4 @@ registry:
     requires_python: ">=3.8"
     deps:
       # https://pypi.org/project/protobuf/
-      - "protobuf"
+      - "protobuf~=4.25"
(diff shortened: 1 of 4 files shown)


pkwarren (approved):

mfridman (approved):

plugins/protocolbuffers/python/v25.0/buf.plugin.yaml
```diff
@@ -14,4 +14,4 @@ registry:
     requires_python: ">=3.8"
     deps:
       # https://pypi.org/project/protobuf/
-      - "protobuf"
+      - "protobuf~=4.25"
```
mfridman, line 17: Which version of protocolbuffers/python is comaptible with protobuf >=5?

mfridman: Never mind, this is irrelevant. I double-checked what we're doing and this makes sense.

PR #1285, opened 2024-06-10
Update docker images for builds

pkwarren:


.github/docker/Dockerfile.python
```diff
@@ -1,3 +1,3 @@
-FROM python:3.12.3-alpine3.19
+FROM python:3.11.9-alpine3.20
```
pkwarren, line 1: Reverting this back to Python 3.11 to keep in sync with the distroless image:
```
$ docker run --rm -it gcr.io/distroless/python3-debian12:latest --version
Python 3.11.2
```

.github/docker/Dockerfile.swift
```diff
@@ -1,3 +1,3 @@
-FROM swift:5.10.1-jammy
+FROM swift:5.10.1-bookworm
```
pkwarren, line 1: Switched to build on the same version that we run on. This doesn't appear to have an impact though since we link the plugins statically.

mfridman (approved):

PR #1291, opened 2024-06-11
Reduce image size and layers in es plugins

pkwarren:


.github/docker/Dockerfile.node
```diff
@@ -1,3 +1,3 @@
-FROM node:20.14.0-alpine3.20
+FROM node:20.14.0-bookworm
```
pkwarren, line 1: Switched to build images using bookworm just so we have the same version of everything (distroless is Debian 12 bookworm).

plugins/bufbuild/es/v1.3.0/Dockerfile
```diff
@@ -1,11 +1,21 @@
-# syntax=docker/dockerfile:1.4
-FROM node:18.16.1-alpine3.18 AS build
+# syntax=docker/dockerfile:1.7
+FROM node:18.20.3-bookworm AS build
```
pkwarren, line 2: Kept the plugins that were on Node 18 on the same version but bumped images to latest security/bugfix releases.

plugins/bufbuild/es/v1.8.0/package.json
```diff
@@ -5,6 +5,6 @@
     "@bufbuild/protoc-gen-es": "1.8.0"
   },
   "devDependencies": {
-    "esbuild": "0.20.2"
+    "esbuild": "^0.21.5"
```
stefanvanburen, line 8: we'll just have to bump these ourselves when esbuild hits `0.22.*`, right?

stefanvanburen (approved):
lgtm

internal/docker/build.go
```diff
@@ -67,13 +68,24 @@ func Build(
...
 	commonArgs = append(commonArgs, args...)
-	f, err := os.Open(filepath.Join(filepath.Dir(plugin.Path), "Dockerfile"))
-	if err != nil {
-		return nil, err
+	buildArgs := slices.Concat(commonArgs, []string{
+		"-t", imageName,
+		filepath.Dir(plugin.Path),
+	})
+	cmd := exec.CommandContext(ctx, "docker", buildArgs...)
+	// Set file modification times to bust Docker cache for local files
```
pkwarren, line 76: This is a pretty ugly hack - I'm not able to reproduce it in local builds (only in CI). Verified that this does appear to fix the build and test failures in this PR however.

An alternative might be to create a temporary directory and copy all the Docker context files to it prior to build, but just setting the mtime does the trick too.

stefanvanburen (approved):

internal/docker/build.go
mfridman: Would this help us in any way? Should we add this regardless?

https://docs.docker.com/build/ci/github-actions/reproducible-builds/

pkwarren: I don't believe that will help in this case. I think the `COPY` command is getting tripped up because the `package.json` and `package-lock.json` files have identical mtime (due to git checkout) and sizes and thinking it can re-use the local cache - https://github.com/moby/buildkit/issues/4817 looks related.

The source epoch would allow our images to be more reproducible across builds which is a good thing but perhaps something we should tweak outside of this change.

mfridman (approved):

PR #1295, opened 2024-06-12
Backport Python `protobuf` compatible versions to pyi v25 versions

stefanvanburen:
We backported these to protocolbuffers/python in #1266, but should also specify them here.

### DIFF
--- a/plugins/protocolbuffers/pyi/v25.0/buf.plugin.yaml
+++ b/plugins/protocolbuffers/pyi/v25.0/buf.plugin.yaml
@@ -14,5 +14,5 @@ registry:
     requires_python: ">=3.8"
     deps:
       # https://pypi.org/project/protobuf/
-      - "protobuf"
+      - "protobuf~=4.25"
       - "types-protobuf"
--- a/plugins/protocolbuffers/pyi/v25.3/buf.plugin.yaml
+++ b/plugins/protocolbuffers/pyi/v25.3/buf.plugin.yaml
@@ -14,5 +14,5 @@ registry:
     requires_python: ">=3.8"
     deps:
       # https://pypi.org/project/protobuf/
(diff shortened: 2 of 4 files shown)


plugins/protocolbuffers/pyi/v25.0/buf.plugin.yaml
```diff
@@ -14,5 +14,5 @@ registry:
     requires_python: ">=3.8"
     deps:
       # https://pypi.org/project/protobuf/
-      - "protobuf"
+      - "protobuf~=4.25"
       - "types-protobuf"
```
stefanvanburen, line 18: FWIW, types-protobuf seemingly follows a similar versioning scheme, but it seems like the PRs hang out a lot longer in typeshed: https://github.com/python/typeshed/pull/12026. So if we wanted to specify a range, we'd probably have to wait until those upstream PRs land.

plugins/protocolbuffers/pyi/v25.3/buf.plugin.yaml
```diff
@@ -14,5 +14,5 @@ registry:
     requires_python: ">=3.8"
     deps:
       # https://pypi.org/project/protobuf/
-      - "protobuf"
+      - "protobuf~=4.25"
```
pkwarren, line 17: I've pulled this change in https://github.com/bufbuild/plugins/pull/1287/commits/f33db4d8e03ad630d2188c1fada476ba08c2162d which will hopefully land today (it has some dependencies on the CLI to get a green build). I'd prefer to do this together so we only bump the revision once.

stefanvanburen:
Coördinated with @pkwarren to get these changes into the image-size reducing PRs separately a la #1291, so closing this one out.

PR #1301, opened 2024-06-14
Reduce image layers in protobuf-ts plugins

pkwarren:


mfridman (approved):
Really glad we had the concept of revision 😌.

PR #1302, opened 2024-06-14
Reduce image layers in connectrpc/es plugins

pkwarren:


mfridman (approved):
Thanks for looking into this 🎉

PR #1315, opened 2024-06-24
Temporarily disable roadrunner-server/grpc

mfridman:


plugins/community/roadrunner-server-php-grpc/source.yaml
```diff
@@ -1,4 +1,8 @@
 source:
+  # Temporarily disabled until someone requests a new version or until
+  # we add the feature to update versions on a less frequent basis, see:
+  # https://github.com/bufbuild/plugins/issues/1039
+  disabled: false
```
DMarby, line 5: should this be `disabled: true`? 😅

DMarby (approved):

plugins/community/roadrunner-server-php-grpc/source.yaml
mfridman: yes...

PR #1321, opened 2024-06-27
Fix python v26.1 plugin dependencies

pkwarren:
Update python v26.1 plugin dependencies to pull in the same or later version of the protobuf library. While we're rebuilding plugins, reduce additional layers in the Python v26.1 plugins and add a dependency on pyi plugin to match updates in later versions.

Ref: https://github.com/bufbuild/plugins/pull/1319

### DIFF
--- a/plugins/protocolbuffers/pyi/v26.1/buf.plugin.yaml
+++ b/plugins/protocolbuffers/pyi/v26.1/buf.plugin.yaml
@@ -14,5 +14,5 @@ registry:
     requires_python: ">=3.8"
     deps:
       # https://pypi.org/project/protobuf/
-      - "protobuf~=5.26"
-      - "types-protobuf"
+      - "protobuf~=5.26.1"
+      - "types-protobuf~=5.26"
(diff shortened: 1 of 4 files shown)


plugins/protocolbuffers/pyi/v26.1/buf.plugin.yaml
```diff
@@ -14,5 +14,5 @@ registry:
     requires_python: ">=3.8"
     deps:
       # https://pypi.org/project/protobuf/
-      - "protobuf~=5.26"
-      - "types-protobuf"
+      - "protobuf~=5.26.1"
+      - "types-protobuf~=5.26"
```
stefanvanburen, line 18: guess we could have done this in the v27 plugins now that there's a release for them, but probably not worth another rev this time.

stefanvanburen (approved):

plugins/protocolbuffers/pyi/v26.1/buf.plugin.yaml
pkwarren: I paused the release workflow as they haven't gone out yet but I can update v27.1 at least with this.

pkwarren: https://github.com/bufbuild/plugins/pull/1322

PR #1332, opened 2024-07-03
Bump go dependency to 1.22.5

pkwarren:


pkwarren:
Dependabot determined the latest version is 1.23rc1 (#1331) and so we need to manually update this.

PR #1344, opened 2024-07-12
Add `std` feature to prost dependency

stefanvanburen:
Since we're not supporting `no_std` crates, we should enable the `std` feature by default. (This happens to be the same as `default_features` currently, but that may change in the future.)


stefanvanburen:
[Related public slack post](https://bufbuild.slack.com/archives/CRZ680FUH/p1720766722923389). Ultimately, not all crates need the `std` features, but this doesn't (currently) add any additional dependencies to the graph, just flips some codegen bits so you don't need to manually `cargo add prost@<matching-version> --features std`.

mfridman (approved):

plugins/community/neoeinstein-prost/v0.3.1/buf.plugin.yaml
```diff
@@ -16,6 +16,7 @@ registry:
       - name: "prost"
         req: "0.12.3"
         default_features: false
+        features: ["std"]
```
pkwarren, line 19: Is this difference between this and setting `default_features: true` that it would enable the `derive` feature too?

plugins/community/neoeinstein-prost/v0.3.1/buf.plugin.yaml
```diff
@@ -16,6 +16,7 @@ registry:
...
       # prost-types is necessary for any module using the WKTs.
       - name: "prost-types"
```
pkwarren, line 21: Should we do the same here? https://docs.rs/prost-types/0.13.1/prost_types/#feature-flags

plugins/community/neoeinstein-prost/v0.3.1/buf.plugin.yaml
stefanvanburen: yeah, that's the difference. Not sure if we want to enable `derive` or not. My vague understanding (from the [README](https://docs.rs/prost/0.12.3/prost/index.html#serializing-existing-types)) is that it's more for user-generated types, but OTOH the custom `derive` macro is added to the generated code.

Let's maybe just add it (by using `default_features: true`) — it's a single extra crate with no dependencies.

pkwarren (approved):

PR #1349, opened 2024-07-15
Add NuGet repository info to csharp plugins

pkwarren:


plugins/grpc/csharp/v1.65.0/buf.plugin.yaml
```diff
@@ -9,3 +9,11 @@ output_languages:
   - csharp
 spdx_license_id: Apache-2.0
 license_url: https://github.com/grpc/grpc/blob/v1.65.0/LICENSE
+registry:
+  nuget:
+    target_frameworks:
+      - netstandard2.0
+    deps:
+      - name: Grpc.Net.Common
+        version: 2.63.0
```
pkwarren, line 18: Unclear how to determine which version of these packages is required by the protoc plugin, so I'm starting w/ the latest stable version here.

.github/workflows/arm64.yml
```diff
@@ -37,10 +37,16 @@ jobs:
...
+    - name: Get buf version
+      shell: bash
+      run: |
+        echo BUF_VERSION=$(go mod edit -json | jq -r ".Require.[] | select(.Path == \"github.com/bufbuild/buf\") | .Version") >> $GITHUB_ENV
     - name: Install buf cli
-      uses: bufbuild/buf-setup-action@v1
-      with:
-        github_token: ${{ github.token }}
+      run: |
+        go install github.com/bufbuild/buf/cmd/buf@${BUF_VERSION}
```
mfridman, line 46: It seems like a lot of effort 🤷 . Could keep it simple and `@main` and then `go get`. 

But I get what you;re trying to do.

mfridman: Ah, it's targeting a branch.

pkwarren: I wanted to get this in place so going forward we'd only have a single version to track if we need to upgrade to a newer dev version. I don't like pulling in `main` as it isn't a reproducible build.

mfridman: Gotcha, disregard my initial comment. This is actually a good idea 😄

mfridman (approved):

stefanvanburen (approved):

PR #1379, opened 2024-07-25
Fix gRPC gateway plugin patch for aliased import path for generated SDK

oliversun9:
This fixes the bug where the deprecation message from the package at the incorrect location would say this package has moved to `buf.build/gen/go/owner/module/protocolbuffers/go/service1/v1/service1v1gateway` while it should be `buf.build/gen/go/owner/module/grpc-ecosystem/gateway/service1/v1/service1v1gateway`. 

Cause: The patch was incorrectly taking `go_package`, which is the location of code generated by `protocolbuffers/go`, as the location of a package generated by this plugin.

[Actual code diff](https://github.com/oliversun9/grpc-gateway/compare/osn/alias...osn/alias-fix).


plugins/grpc-ecosystem/gateway/v2.20.0/separate_pkg_additional_imports.patch
```diff
@@ -237,7 +237,9 @@ index 62766254..8084ff74 100644
 +		//
 +		// If gateway has a new major version, we should then stop generating at the wrong path.
 +		aliasedPackage := &descriptor.GoPackage{
-+			Path:  goPkg.Path,
++			// When generating for generated SDK, the original goPkg points to code generated by "protocolbuffers/go",
++			// but we are aliasing to a package generated by "grpc-ecosystem/gateway".
++			Path:  strings.Replace(goPkg.Path, "protocolbuffers/go", "grpc-ecosystem/gateway", 1),
```
pkwarren, line 242: I think we already have constants for protocolbuffers/go and grpc/go in internal/descriptor/buf_build.go. It would be good to centralize all of those in one place.

We should also consider changing the replace to be more specific `/protocolbuffers/go` -> `/grpc-ecosystem/gateway`.

oliversun9: Fixed in [f8c1369](https://github.com/bufbuild/plugins/pull/1379/commits/f8c1369107878d8cff05d84d24d451b5e634feba) and [868f508](https://github.com/bufbuild/plugins/pull/1379/commits/868f50836f39c185b2f9a6c97a7a3107bb4c9f84)

pkwarren (approved):

PR #1389, opened 2024-07-29
Automatically replace docker syntax to the desired version when fetching

oliversun9:
This would automatically replace `# syntax=docker/dockerfile:1.7` at the top of the Dockerfile to 1.9.

### DIFF
--- a/internal/cmd/fetcher/main.go
+++ b/internal/cmd/fetcher/main.go
@@ -25,9 +25,11 @@ import (
 )
 
 var (
-	bazelDownloadRegexp = regexp.MustCompile(`bazelbuild/bazel/releases/download/[^/]+/bazel-[^-]+-linux`)
-	bazelImageName      = "gcr.io/bazel-public/bazel"
-	errNoVersions       = errors.New("no versions found")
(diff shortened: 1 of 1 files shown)


internal/cmd/fetcher/main.go
```diff
@@ -25,9 +25,11 @@ import (
...
 
 var (
-	bazelDownloadRegexp = regexp.MustCompile(`bazelbuild/bazel/releases/download/[^/]+/bazel-[^-]+-linux`)
-	bazelImageName      = "gcr.io/bazel-public/bazel"
-	errNoVersions       = errors.New("no versions found")
+	bazelDownloadRegexp   = regexp.MustCompile(`bazelbuild/bazel/releases/download/[^/]+/bazel-[^-]+-linux`)
+	bazelImageName        = "gcr.io/bazel-public/bazel"
+	errNoVersions         = errors.New("no versions found")
+	dockerSyntaxPrefix    = "# syntax=docker/dockerfile:"
+	preferredDockerSyntax = "1.9"
```
mfridman, line 32: Not quite, we should always be able to get this value from this file, which itself is kept up-to-date via dependabot.

https://github.com/bufbuild/plugins/blob/main/.github/docker/Dockerfile.dockerfile

mfridman (approved):
Thank you 🎉

PR #1397, opened 2024-08-01
Enable default_features in tonic dependency

stefanvanburen:
Spending more time with the generated code, it seems that all of the current default_features are required for code generation - we ought to flip this to `true` so that users don't need to.

### DIFF
--- a/plugins/community/neoeinstein-tonic/v0.4.0/buf.plugin.yaml
+++ b/plugins/community/neoeinstein-tonic/v0.4.0/buf.plugin.yaml
@@ -16,8 +16,7 @@ registry:
       - name: "tonic"
         req: "0.11.0"
         # https://github.com/hyperium/tonic/blob/v0.11.0/tonic/Cargo.toml#L29
-        # The generated code does not require any of the default features of tonic.
-        default_features: false
+        default_features: true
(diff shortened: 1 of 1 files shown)


pkwarren (approved):

stefanvanburen:
Quickly explaining the different [default features](https://github.com/hyperium/tonic/blob/v0.11.0/tonic/Cargo.toml#L29) and why we should enable them:

* Both generated [client][1] and [server][2] code add an unconditional `use` on `tonic::codegen`, so we need the `codegen` feature to pick up that dependency
* We aren't specifying the [`no_transport` option](https://github.com/neoeinstein/protoc-gen-prost/blob/protoc-gen-tonic-v0.4.0/protoc-gen-tonic/README.md#options), so we should enable `transport`
* Without doing custom overrides of the codec, we end up with the ProstCodec ([related code](https://github.com/hyperium/tonic/blob/v0.11.0/tonic-build/src/prost.rs#L64)) being generated by default, meaning that we need the `prost` feature

[1]: https://github.com/hyperium/tonic/blob/v0.11.0/tonic-build/src/client.rs#L55-L56
[2]: https://github.com/hyperium/tonic/blob/v0.11.0/tonic-build/src/server.rs#L106

pkwarren (approved):

pkwarren (approved):

PR #1428, opened 2024-08-14
Validate maven dependency availability

pkwarren:
Prior to releasing a plugin supporting Maven generated SDKs, we should ensure that all dependencies declared in the buf.plugin.yaml file are available.


mfridman (approved):
I do really wish we could bake these deps _into_ the plugin so it's self-contained.

stefanvanburen (approved):

PR #1429, opened 2024-08-14
Add `dotnet restore` step for NuGet enabled plugins

stefanvanburen:
This adds a `dotnet restore` step to the docker build for all NuGet enabled plugins, placing the output of the nuget packages to `/nuget`.


stefanvanburen:
Would like to do some more internal testing, but open to comments on this approach. Added a test that should verify that the buf.plugin.yaml + build.csproj are in sync.

plugins/grpc/csharp/v1.65.0/Dockerfile
```diff
@@ -17,10 +17,16 @@ WORKDIR /build/grpc
 RUN bazel build //src/compiler:grpc_plugin_support
 RUN bazel build //src/compiler:grpc_csharp_plugin.stripped
 
+FROM mcr.microsoft.com/dotnet/sdk:6.0.424-bookworm-slim@sha256:39f73995c6d4f64f1a81f4ec42617cb71a797a23da93a5b0109ab06e5094fdae AS dotnetrestore
```
pkwarren, line 20: We should probably add this image to .github/docker so we can keep it up to date with the desired version of the dotnet SDK.

plugins/grpc/csharp/v1.65.0/Dockerfile
```diff
@@ -17,10 +17,16 @@ WORKDIR /build/grpc
...
+WORKDIR /build
+COPY ./build.csproj /build
```
pkwarren, line 22: Can do a `--link` here.

tests/plugins_test.go
```diff
@@ -298,6 +299,71 @@ func TestMavenDependencies(t *testing.T) {
...
+	allPlugins := loadAllPlugins(t)
+	plugins := loadFilteredPlugins(t)
+	for _, p := range plugins {
+		if p.Registry.Nuget == nil {
+			continue
+		}
+		t.Run(p.String(), func(t *testing.T) {
+			t.Parallel()
+			// We require all NuGet-enabled plugins to have a build.csproj file to load plugin dependencies.
+			buildCsproj, err := os.Open(filepath.Join(filepath.Dir(p.Path), "build.csproj"))
```
pkwarren, line 332: Can do an `os.ReadFile` here to remove some boilerplate.

plugins/grpc/csharp/v1.65.0/Dockerfile
stefanvanburen: Added a Dockerfile in https://github.com/bufbuild/plugins/pull/1429/commits/88f387bcad94694278536072a84812a31d615ed8 — assuming it'll still pick up updates with the specific sha256 specified?

stefanvanburen:
Ok, split out the updates here into #1435, #1436 and #1437 to take care of the plugin bazel build times 😄.

DMarby (approved):
LGTM, but would like another approval from @mfridman and/or @pkwarren since they're more familiar with this repo then I am 😅

stefanvanburen:
Bumped the image version for running `dotnet restore` to dotnet/sdk:8; confirmed that it behaves in the same way, so we can use the latest LTS version instead of 6.

plugins/grpc/csharp/v1.65.4/Dockerfile
```diff
@@ -17,10 +17,16 @@ WORKDIR /build/grpc
...
+FROM mcr.microsoft.com/dotnet/sdk:8.0.401-bookworm-slim@sha256:8c6beed050a602970c3d275756ed3c19065e42ce6ca0809f5a6fcbf5d36fd305 AS dotnetrestore
+WORKDIR /build
+COPY --link ./build.csproj /build/build.csproj
+RUN mkdir /nuget && dotnet restore --packages /nuget
+
 FROM gcr.io/distroless/cc-debian12:latest@sha256:3b75fdd33932d16e53a461277becf57c4f815c6cee5f6bc8f52457c095e004c8 AS base
 
 FROM scratch
 COPY --link --from=base / /
+COPY --link --from=dotnetrestore --chmod=0755 /nuget /nuget
```
pkwarren, line 29: We should leave the chmod off here - all the files within don't need their permissions changed.

pkwarren (approved):
One comment but otherwise looks great.

pkwarren (approved):

plugins/grpc/csharp/v1.65.4/Dockerfile
stefanvanburen: Dropped from all affected Dockerfiles in https://github.com/bufbuild/plugins/pull/1429/commits/65378b84d700c5d3f89cf7e579e43c92c41edd6f — I'll update the other PRs as well.

PR #1431, opened 2024-08-15
Update to go 1.23

pkwarren:


mfridman (approved):
Should we use 'latest'?

PR #1456, opened 2024-08-27
Add `Grpc.Net.Client` dependency to grpc/csharp

stefanvanburen:
Matching with the `Grpc.Net.Common` dependency version.

### DIFF
--- a/plugins/grpc/csharp/v1.65.0/build.csproj
+++ b/plugins/grpc/csharp/v1.65.0/build.csproj
@@ -4,6 +4,7 @@
   </PropertyGroup>
   <ItemGroup>
     <PackageReference Include="Grpc.Net.Common" Version="2.63.0" />
+    <PackageReference Include="Grpc.Net.Client" Version="2.63.0" />
     <PackageReference Include="Google.Protobuf" Version="3.26.1" />
   </ItemGroup>
 </Project>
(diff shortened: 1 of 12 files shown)


perezd (approved):

plugins/grpc/csharp/v1.65.0/build.csproj
```diff
@@ -4,6 +4,7 @@
   </PropertyGroup>
   <ItemGroup>
     <PackageReference Include="Grpc.Net.Common" Version="2.63.0" />
+    <PackageReference Include="Grpc.Net.Client" Version="2.63.0" />
```
stefanvanburen, line 7: technically these don't need to be here as they aren't used in compilation, but we're currently testing to ensure that they match. Ideally we'd separate out this dep from the others, but .. 🤷‍♂️? Shouldn't matter much.

stefanvanburen:
(I probably need to split up this PR into a couple, as the build times for these bazel plugins is long.)

stefanvanburen:
Going to just add this dep moving forward, as grpc/csharp is bumping to 1.66.1 today: https://github.com/bufbuild/plugins/pull/1455/commits/f02c183b055fe840cba9942c99cd3b7a1b838b51.

PR #1561, opened 2024-10-25
Mark latest versions of disabled plugins as deprecated

mfridman:
This PR sets `deprecated: true` for disabled plugins. This will help fix issues where some clusters we manage have outdated information.


plugins/bufbuild/connect-es/v0.13.0/buf.plugin.yaml
```diff
@@ -17,3 +17,4 @@ registry:
         version: ^0.13.0
 spdx_license_id: Apache-2.0
 license_url: https://github.com/bufbuild/connect-es/blob/v0.13.0/LICENSE
+deprecated: true
```
pkwarren, line 20: It would be good to update `cmd/latest-plugins/main.go` method `latestNonDeprecatedPlugins` to use this to filter out deprecated plugins.

pkwarren (approved):
If we ever were to publish an old version of a deprecated version for any reason, would that undo this change since it is only being updated for the latest version?

plugins/community/chrusty-jsonschema/source.yaml
```diff
@@ -1,4 +1,5 @@
 source:
+  disabled: true
```
pkwarren, line 2: We need to add this plugin to `isDeprecated` in the latest-plugins command.

mfridman:
> If we ever were to publish an old version of a deprecated version for any reason, would that undo this change since it is only being updated for the latest version?

Yes, although this feels like a _very_ small edge case. The alternative is to do this for all versions, but that'll cause revision bumps and that's slightly worse.

I think at some point we should prune deprecated/disabled plugins wholesale.

pkwarren:
> Yes, although this feels like a _very_ small edge case. The alternative is to do this for all versions, but that'll cause revision bumps and that's slightly worse.

Agree - alternatively we could only pay attention to the deprecated attribute if the plugin being installed is the latest version.

mfridman:
> Agree - alternatively we could only pay attention to the deprecated attribute if the plugin being installed is the latest version.

I think I like the property of latest writer wins, regardless of version. We just have to be mindful of that. And even then, worst case you push a new revision.

PR #1597, opened 2024-11-20
Deprecate connectrpc es and query-es

mfridman:
This PR marks the following plugins as deprecated with a new description:

- https://buf.build/connectrpc/es
- https://buf.build/connectrpc/query-es

And updates the description for:

- https://buf.build/bufbuild/connect-es (previously deprecated)
- https://buf.build/bufbuild/connect-query (previously deprecated)
- https://github.com/bufbuild/protobuf-es

After this is merged/released, I'll make them as `disabled: true` in the source. I don't recall if we use that for any logic outside the fetcher.


stefanvanburen (commented):
seems good, just not sure about the markdown in the description as I didn't see other descriptions w/ markdown in them?

stefanvanburen (approved):

PR #1638, opened 2024-12-19
Add connect-dart plugin

srikrsna-buf:
Add connect-dart plugin


plugins/connectrpc/dart/source.yaml
```diff
@@ -0,0 +1,3 @@
+source:
+  dart_flutter:
+    name: connectrpc
```
mfridman, line 3: Double-checking, I presume we will be publishing to pub.dev?

mfridman: In other words, is this the correct place to look for new releases instead of just github?

plugins/connectrpc/dart/v0.1.0/buf.plugin.yaml
```diff
@@ -0,0 +1,12 @@
...
+plugin_version: v0.1.0
+source_url: https://github.com/connectrpc/connect-dart
+integration_guide_url: https://connectrpc.com/docs/dart/getting-started
+description: Generates client stubs for connectrpc. Compatible with the gRPC, gRPC-Web, and Connect RPC protocols.
+deps:
+  - plugin: buf.build/protocolbuffers/dart:v21.1.2
+output_languages:
+  - dart
+spdx_license_id: Apache-2.0
+license_url: https://github.com/connectrpc/connect-dart/blob/LICENSE
```
mfridman, line 12:
```suggestion
license_url: https://github.com/connectrpc/connect-dart/blob/v0.1.0/LICENSE
```

plugins/connectrpc/dart/v0.1.0/buf.plugin.yaml
```diff
@@ -0,0 +1,12 @@
+version: v1
+name: buf.build/connectrpc/dart
+plugin_version: v0.1.0
+source_url: https://github.com/connectrpc/connect-dart
+integration_guide_url: https://connectrpc.com/docs/dart/getting-started
+description: Generates client stubs for connectrpc. Compatible with the gRPC, gRPC-Web, and Connect RPC protocols.
```
mfridman, line 6: Is this description accurate for the `connec-`dart implementation, it supports gRPC-Web?

plugins/connectrpc/dart/v0.1.0/Dockerfile
```diff
@@ -0,0 +1,16 @@
+# syntax=docker/dockerfile:1.4
```
mfridman, line 1:
```suggestion
# syntax=docker/dockerfile:1.12
```

plugins/connectrpc/dart/v0.1.0/Dockerfile
```diff
@@ -0,0 +1,16 @@
...
+FROM dart:3.5.4-sdk AS build
```
mfridman, line 2: What is the best way to know which dart version we should support?

Asking because there's an upstream issue where the directive is still at 2.x

https://github.com/google/protobuf.dart/issues/936

plugins/connectrpc/dart/source.yaml
srikrsna-buf: Updated in https://github.com/bufbuild/plugins/pull/1638/commits/a166d0d9ba4ca52046b773e53d95a6e57e933fc0

plugins/connectrpc/dart/v0.1.0/Dockerfile
mfridman: Okay great, then we should be able to update the latest base types to the latest dart 3 version. That way we can keep both dart plugins on the same version.

plugins/connectrpc/dart/v0.1.0/Dockerfile
```diff
@@ -0,0 +1,16 @@
+# syntax=docker/dockerfile:1.12
+FROM dart:3.5.4-sdk AS build
+
+WORKDIR /build
+RUN curl -fsSL https://github.com/connectrpc/connect-dart/archive/refs/tags/v0.1.0.tar.gz --compressed -o connect-dart.tar.gz \
```
mfridman, line 5: We typically `RUN git clone --depth 1` instead of fetching the .zip, any reason we have to deviate here from the other Dockerfiles?

mfridman (commented):
lgtm. Just one nit on using git clone instead of curl the .zip for consistency

plugins/connectrpc/dart/v0.1.0/Dockerfile
srikrsna-buf: This is similar to the dart base plugin image, we just swapped the url from pub.dev to github and cd into the directory. I can change it to git clone if that is better suited

mfridman: Would prefer if we used the `git clone` approach.

mfridman (approved):

PR #1689, opened 2025-01-31
Stop updating Bazel version in builds

pkwarren:
We now use Bazel with bazelisk and pin exact versions (either the version specified in .bazelversion or in an env var). Remove Docker image for bazel and code to auto update Bazel in the fetcher.


mfridman (approved):
:relief:

PR #1723, opened 2025-02-19
Add grpc/swift-protobuf

### LINKED ISSUE #1696: Support grpc-swift v2.0.0
In https://github.com/bufbuild/plugins/pull/1695, we've temporarily disabled updates to the grpc-swift plugin. The source code for the plugin is now found in a separate grpc-swift-protobuf project, which is versioned at v1.0.0.

We have a few options:

1. Continue tracking updates to grpc-swift but change the Docker build to ship grpc-swift-protobuf v1.0.0 artifacts. We'd have to manually track future updates and this wouldn't be great as the version of the plugin wouldn't match the version of the source used to build it. For example, if grpc-swift-protobuf bumped to v1.0.1 but grpc-swift stayed at v2.0.0, the best we could do is rebuild the docker image to push a new revision of grpc-swift.
2. Stop updates to current grpc-swift plugin. Create a new plugin to track grpc-swift-protobuf.

In either case, we'll need to update the Swift registry dependencies and the URLs for the project in buf.plugin.yaml.

stefanvanburen:
Currently at v1.0.0.

Resolves #1696.

### DIFF
--- a/plugins/grpc/swift-protobuf/v1.0.0/buf.plugin.yaml
+++ b/plugins/grpc/swift-protobuf/v1.0.0/buf.plugin.yaml
@@ -1,0 +1,23 @@
+version: v1
+name: buf.build/grpc/swift-protobuf
+plugin_version: v1.0.0
+source_url: https://github.com/grpc/grpc-swift-protobuf
+description: Generates Swift client and server stubs for the gRPC framework.
+deps:
+  # https://github.com/grpc/grpc-swift-protobuf/blob/1.0.0/Package.swift#L42
+  - plugin: buf.build/apple/swift:v1.28.2
+output_languages:
+  - swift
+registry:
+  swift:
+    deps:
+      - source: https://github.com/grpc/grpc-swift-protobuf.git
+        package: grpc-swift-protobuf
+        # https://github.com/grpc/grpc-swift-protobuf/blob/1.0.0/Package.swift#L22
+        products: [ GRPCProtobuf ]
+        version: 1.0.0
+  opts:
+    - Visibility=Public
+    - FileNaming=PathToUnderscores
+spdx_license_id: Apache-2.0
+license_url: https://github.com/grpc/grpc-swift-protobuf/blob/1.0.0/LICENSE
--- a/plugins/grpc/swift-protobuf/v1.0.0/buf.plugin.yaml
+++ b/plugins/grpc/swift-protobuf/v1.0.0/buf.plugin.yaml
@@ -2,6 +2,7 @@ version: v1
 name: buf.build/grpc/swift-protobuf
 plugin_version: v1.0.0
 source_url: https://github.com/grpc/grpc-swift-protobuf
+integration_guide_url: https://swiftpackageindex.com/grpc/grpc-swift/2.0.0/documentation/grpccore
 description: Generates Swift client and server stubs for the gRPC framework.
 deps:
   # https://github.com/grpc/grpc-swift-protobuf/blob/1.0.0/Package.swift#L42
--- a/plugins/grpc/swift-protobuf/v1.0.0/buf.plugin.yaml
+++ b/plugins/grpc/swift-protobuf/v1.0.0/buf.plugin.yaml
@@ -16,6 +16,14 @@ registry:
         # https://github.com/grpc/grpc-swift-protobuf/blob/1.0.0/Package.swift#L22
         products: [ GRPCProtobuf ]
         version: 1.0.0
+        # https://github.com/grpc/grpc-swift-protobuf/blob/1.0.0/Package.swift#L47
+        swift_versions: [ ".v6" ]
+        # https://github.com/grpc/grpc-swift-protobuf/blob/1.0.0/Package.swift#L123-L126
+        platforms:
+          macos: v15
+          ios: v18
+          tvos: v18
+          watchos: v11
   opts:
     - Visibility=Public
     - FileNaming=PathToUnderscores
--- a/plugins/grpc/swift-protobuf/v1.0.0/buf.plugin.yaml
+++ b/plugins/grpc/swift-protobuf/v1.0.0/buf.plugin.yaml
@@ -5,8 +5,7 @@ source_url: https://github.com/grpc/grpc-swift-protobuf
 integration_guide_url: https://swiftpackageindex.com/grpc/grpc-swift-protobuf/1.0.0/documentation/grpcprotobuf
 description: Generates Swift client and server stubs for the gRPC framework.
 deps:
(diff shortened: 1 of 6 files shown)


plugins/grpc/swift-protobuf/v1.0.0/buf.plugin.yaml
```diff
@@ -0,0 +1,23 @@
+version: v1
+name: buf.build/grpc/swift-protobuf
+plugin_version: v1.0.0
+source_url: https://github.com/grpc/grpc-swift-protobuf
```
stefanvanburen, line 4: no `integration_guide`, currently: the one [linked to in the old repo](https://github.com/bufbuild/plugins/blob/c6e380cb6f6ff50a05d3101396d5e5e66dec51b7/plugins/grpc/swift/v1.24.2/buf.plugin.yaml#L5) didn't look valid anymore. We could maybe point over to the [Swift Package Index docs](https://swiftpackageindex.com/grpc/grpc-swift-protobuf/1.0.0/documentation/grpcprotobuf)?

plugins/grpc/swift-protobuf/v1.0.0/buf.plugin.yaml
```diff
@@ -0,0 +1,23 @@
...
+description: Generates Swift client and server stubs for the gRPC framework.
+deps:
+  # https://github.com/grpc/grpc-swift-protobuf/blob/1.0.0/Package.swift#L42
+  - plugin: buf.build/apple/swift:v1.28.1
```
pkwarren, line 8: Should we update this to the latest 1.28.2?

tests/testdata/buf.build/grpc/swift-protobuf/v1.0.0/eliza/plugin.sum
```diff
@@ -0,0 +1 @@
+h1:Ouy7fwcPAu9owIHJWt8sSknVb3j9zb27aNgesYMbWrE=
```
stefanvanburen, line 1: took a peek at the generated code and looked right to me, but would be good to double-check.

plugins/grpc/swift-protobuf/v1.0.0/buf.plugin.yaml
```diff
@@ -0,0 +1,23 @@
+version: v1
+name: buf.build/grpc/swift-protobuf
+plugin_version: v1.0.0
+source_url: https://github.com/grpc/grpc-swift-protobuf
+description: Generates Swift client and server stubs for the gRPC framework.
```
pkwarren, line 5: Do we want an integration_guide_url? Something like: https://swiftpackageindex.com/grpc/grpc-swift/2.0.0/documentation/grpccore

stefanvanburen: works for me, https://github.com/bufbuild/plugins/pull/1723/commits/71e437ea3297b90cde4cea6c73276b932a0d1835

plugins/grpc/swift-protobuf/v1.0.0/buf.plugin.yaml
stefanvanburen: resolved in https://github.com/bufbuild/plugins/pull/1723/commits/71e437ea3297b90cde4cea6c73276b932a0d1835

plugins/grpc/swift-protobuf/v1.0.0/buf.plugin.yaml
stefanvanburen: Do we try to normally match the latest patch release? I dropped it down in https://github.com/bufbuild/plugins/pull/1723/commits/c6e380cb6f6ff50a05d3101396d5e5e66dec51b7 to match the dep in Package.swift; happy to match the latest though.

pkwarren: As long as it is compatible we typically bump the deps to the latest version.

plugins/grpc/swift-protobuf/v1.0.0/buf.plugin.yaml
```diff
@@ -0,0 +1,32 @@
...
+        version: 1.0.0
+        # https://github.com/grpc/grpc-swift-protobuf/blob/1.0.0/Package.swift#L47
+        swift_versions: [ ".v6" ]
+        # https://github.com/grpc/grpc-swift-protobuf/blob/1.0.0/Package.swift#L123-L126
+        platforms:
+          macos: v15
+          ios: v18
+          tvos: v18
+          watchos: v11
+  opts:
```
rebello95, line 28: Just calling out that these are (assumed deliberate) deviations from the defaults here: https://swiftpackageindex.com/grpc/grpc-swift-protobuf/1.0.0/documentation/grpcprotobuf/generating-stubs#Generator-options

rebello95 (commented):
yaml looks good to me, couple callouts

plugins/grpc/swift-protobuf/v1.0.0/buf.plugin.yaml
stefanvanburen: yeah, I had copied these options over from [the latest version of grpc/swift](https://github.com/bufbuild/plugins/blob/551404ca2f2714eee90c0cd2c1bfd714bccef6a2/plugins/grpc/swift/v1.24.2/buf.plugin.yaml#L18-L20), but I ought to check that they work the same way in the new plugin. Will check on that before merge, thanks!

pkwarren (approved):

plugins/grpc/swift-protobuf/v1.0.0/buf.plugin.yaml
stefanvanburen: https://github.com/bufbuild/plugins/pull/1723/commits/bf7eb42435b5eb78c8bc4b1e0c9536d5cc78e7d9, thanks!

pkwarren (approved):

plugins/grpc/swift-protobuf/v1.0.0/buf.plugin.yaml
stefanvanburen: Just took a quick peek: I believe we need `Visibility=Public` since we're publishing a package intended to be consumed by other packages. In terms of `FileNaming=PathToUnderscore`, I see in apple/swift's plugin docs that [this option can help for publishing with SPM](https://github.com/apple/swift-protobuf/blob/d72aed98f8253ec1aa9ea1141e28150f408cf17f/Documentation/PLUGIN.md?plain=1#L101-L103), which makes sense for what we're doing. (I didn't see any discussion in the [original PR that added those options](https://github.com/bufbuild/plugins/pull/508), so assuming that's the point.)

Thanks for calling out, good to check!

PR #1747, opened 2025-03-07
Add protoc v25.6 plugins

### LINKED ISSUE #1746: v25.6 Protobuf packages for Java/Kotlin
[Protobuf v25.6](https://github.com/protocolbuffers/protobuf/releases/tag/v25.6) is the latest release before the 4.x breaking change for JVM-based languages but it's v25.3 is the latest version available in the plugin registry.

Would it be possible to make the latest version available in the previous major version, please?

We are unfortunately stuck on the 3.x Protobuf release for some JVM services for now, so it would be good to know we have the latest bug fixes included in our generated code.

pkwarren:
The protobuf 26.x release broke compatibility for C++, Java, Kotlin, PHP, Python, and Ruby. Most users have upgraded by now, however it may take some time for Java/Kotlin libraries and ecosystem to be updated to support the latest protobuf-java runtime.

Backfill v25.6 which is the latest release on the 25.x line. This also includes a security fix for the Java runtime in GHSA-735f-pc8j-v9w8.

Fixes #1746.


pkwarren (commented):
The procedure I used to create these was to run the fetcher command (patched locally) to target v25.6. This copied the latest best practices from our newest plugin back to this release. This also makes new functionality like csharp gen SDKs available to this release.

The builds however failed. We needed to further set the `USE_BAZEL_VERSION` env var as the latest release of protoc pins Bazel, but earlier versions didn't.

Then I compared v25.3 to v25.6 of each plugin, copying over fixes to metadata in buf.plugin.yaml. We also had patched a few plugins that were incorrectly advertising support for editions before they supported them.

@mfridman When reviewing take a close look at `buf.plugin.yaml` for each plugin. I believe everything is correct but it is worth double checking.

mfridman:
Okay, I'll take a closer pass at this first thing tomorrow morning.

mfridman (approved):

PR #1776, opened 2025-03-19
Remove changed-files GHA

unmultimedio:
Remove the usage of the `tj-actions/changed-files` GHA.

I will work on trying to remove the usage of `nrwl/nx-set-shas` as well, and standarize the command and logs packages for our `./internal/cmd/changed-plugins` command in subsequent PRs.


.github/workflows/pr.yml
```diff
@@ -34,31 +34,17 @@ jobs:
...
       with:
         fetch-depth: 0
-    - name: Get changed files
-      id: changed-files
-      if: ${{ inputs.plugins == '' }}
-      uses: tj-actions/changed-files@2f7c5bfce28377bc069a65ba478de0a74aa0ca32 # v46.0.1
-      with:
-        files: |
-          plugins/**
-          tests/testdata/buf.build/**
```
unmultimedio, line 44: I'm still missing to handle these testdata, will look at it tomorrow

.github/workflows/pr.yml
```diff
@@ -34,31 +34,17 @@ jobs:
...
-        files_ignore: |
-          **/source.yaml
-        separator: ","
-    - name: Show changed files
-      if: ${{ inputs.plugins == '' }}
-      run: |
-        echo '${{ toJSON(steps.changed-files.outputs) }}'
+    - name: Create local base branch too
+      run: git branch ${{ github.base_ref }} remotes/origin/${{ github.base_ref }}
```
pkwarren, line 38: Do we need to fetch more references first this to work properly?

Have you looked at perhaps using https://cli.github.com/manual/gh_pr_diff instead?

internal/plugin/plugin.go
```diff
@@ -342,9 +344,38 @@ func calculateGitModified(ctx context.Context, pluginYamlPath string) (bool, err
...
-// See https://github.com/tj-actions/changed-files#outputs for more details.
-type changedFiles struct {
-	AnyModified      string   `env:"ANY_MODIFIED"`
-	AllModifiedFiles []string `env:"ALL_MODIFIED_FILES"`
+func getBaseRefFromEnv(ctx context.Context, lookuper envconfig.Lookuper) (string, error) {
+	type env struct {
+		BaseRef string `env:"BASE_REF"`
+	}
+	var e env
+	if err := envconfig.ProcessWith(ctx, &envconfig.Config{
```
pkwarren, line 352: We can probably just switch to reading the env var here without envconfig.

internal/plugin/plugin.go
```diff
@@ -342,9 +344,38 @@ func calculateGitModified(ctx context.Context, pluginYamlPath string) (bool, err
...
+			// Include all files within plugins/*
+			if !strings.HasPrefix(filePath, "plugins/") {
+				return false
+			}
+			// We only care about files in the plugins version dirs, no deeper, which means exactly 5
+			// parts in the path:
+			//   plugins/<owner_name>/<plugin_name>/<version>/*
+			//  | 1     | 2          | 3           | 4       | 5
+			parts := strings.Split(filePath, "/")
+			return len(parts) == 5
```
pkwarren, line 378: Just in case any of the docker context ever comes from subdirs:

```suggestion
			return len(parts) >= 5
```

unmultimedio: Ah ok, so anything within the versions dir, even in subdirs?

internal/plugin/plugin.go
unmultimedio: Didn't want to change too much how it was, but sure I can do that

.github/workflows/pr.yml
unmultimedio: We're fetching everything, see L36 `fetch-depth: 0`, so using `remotes/origin/${{ github.base_ref }}` should give us back the true latest commit in `main`.

Just tested and that's what `changed-files` does:

Current latest `main`: `26a7b20ce13a20bbae880f9f776fc2905893eea1`

New PR check https://github.com/bufbuild/plugins/pull/1777
https://github.com/bufbuild/plugins/actions/runs/13971277551/job/39113476165?pr=1777#step:4:67
```
Retrieving changes between 26a7b20ce13a20bbae880f9f776fc2905893eea1 (main) → d4e44bf2da5f70751d966351bdfdb061f054bed9 (fetch-versions)
```

Old PR check https://github.com/bufbuild/plugins/pull/1778 (on top of an older `main` commit)
https://github.com/bufbuild/plugins/actions/runs/13975499975/job/39128135829?pr=1778#step:4:67
```
Retrieving changes between 26a7b20ce13a20bbae880f9f776fc2905893eea1 (main) → 49425aa1f004a252b38289d1c1ae14e248d30906 (jfigueroa/test-old-pr)
```

They both use the latest in `main` as the `base_ref` to diff against.

I will test that it happens the same here as well.

unmultimedio: Resolved via Slack, we'll consider just latest on the default label 👍🏼

unmultimedio: Tested, the `fetch-depth: 0` fetches the true latest `main`, so when checking out to a local `main` it is the true latest commit:

https://github.com/bufbuild/plugins/actions/runs/14001510879/job/39208691711#step:6:12

```
2025/03/21 21:53:15 git ref main resolves to:
ecccc3ca379a708df1eab3ca6d0964632a3ade94 refs/heads/main
ecccc3ca379a708df1eab3ca6d0964632a3ade94 refs/remotes/origin/main
```

SCR-20250321-oxji

No need for additional fetches

unmultimedio: [Switched](https://github.com/bufbuild/plugins/pull/1776/commits/18d9f6fee68769c2a38fbec890bf5cc5a3d289b4) to use the (already fetched) remote ref, [it works](https://github.com/bufbuild/plugins/actions/runs/14109969921/job/39525834007#step:5:13).

pkwarren (approved):

PR #1807, opened 2025-04-17
Enable new funcorder linter

pkwarren:


.golangci.yml
```diff
@@ -5,7 +5,6 @@ linters:
     - cyclop            # covered by gocyclo
     - depguard          # not needed
     - exhaustruct
-    - funcorder         # may enable in the future if matches preferred style
```
pkwarren, line 8: Verified this does enforce rules that we'd like to include in our coding style.

mfridman (approved):

PR #1808, opened 2025-04-17
Connect-query: Make sure to output extensions for module import style

paul-sachs:
We need to make sure the extensions output by connect-query are included for restrictive module systems (like webpack) to work properly.


plugins/connectrpc/query-es/v2.0.2/buf.plugin.yaml
```diff
@@ -0,0 +1,22 @@
...
+  - typescript
+registry:
+  opts:
+    - import_extension=js
+  npm:
+    import_style: module
+    rewrite_import_path_suffix: connectquery.js
+    deps:
+      - package: '@bufbuild/protobuf'
+        version: ^2.2.3
```
paul-sachs, line 20: The only real change in this version of the plugin:
- Added import_extension=js
- Removed dep on @connectrpc/connect-query

plugins/connectrpc/query-es/v2.0.2/buf.plugin.yaml
```diff
@@ -0,0 +1,22 @@
+version: v1
+name: buf.build/connectrpc/query-es
+plugin_version: v2.0.2
```
mfridman, line 3: Wait, why are you changing this without updating the source?

https://github.com/connectrpc/connect-query-es/releases/tag/v2.0.1

paul-sachs: Because we needed different options for the plugin but the source is identical.

mfridman: This is precisely why we have revisions. So you can make config-related changes without modifying anything about the upstream version.

The version you bumped should match whatever is published upstream.

In this case just make the config changes and it'll bump the revision.

paul-sachs: Ah so my original instinct was correct and i regret my follow up.

mfridman: Hehe, what I'm trying to say is your initial commit was correct. It modified the config and that's it, the rest is handles for you, so there's nothing to dig into.

paul-sachs: Yep, all good. Pushed up my original change.

plugins/connectrpc/query-es/v2.0.1/buf.plugin.yaml
```diff
@@ -10,12 +10,12 @@ output_languages:
...
   - typescript
 registry:
+  opts:
+    - import_extension=js
   npm:
     import_style: module
     rewrite_import_path_suffix: connectquery.js
     deps:
-      - package: '@connectrpc/connect-query'
-        version: ^2.0.1
```
timostamm, line 18: Why remove this dep?

paul-sachs: The generated code doesn't actually depend on it. I think this was a mistake when releasing 2.0.0. Using the output generally means you want connect-query, but this change prevents having multiple versions installed (since you can update connect-query separate from it's plugin).

paul-sachs: At least, if I understand how deps work in plugins. Please correct me if i'm wrong.

paul-sachs: Actually, looking at the output of `npm info buf/connectrpc_eliza.connectrpc_query-es`, i expected it to say `@connectrpc/connect-query` and it didn't... so yeah maybe I am confused about what exactly these deps are used for.

mfridman: These deps go into the `package.json`, so when you npm install it pulls in the required deps. 

I haven't followed connect-query to know if the generated code depends on `@connectrpc/connect-query` or not. But if it doesn't, then probably safe to remove. This is where I'd lean on ya'll to know what deps the generated code requires (or not).

paul-sachs: Ah yeah my confusion was because npm info doesn't list peer deps by default. That makes sense.

mfridman:
Can you run `make test PLUGINS="connectrpc/query-es:v2.0.1"` and check those files in.

PR #1834, opened 2025-05-15
Improve ratelimiting behaviour for GitHub

emcfarlane:
This aims to fix ratelimiting issues seens on builds. We switch the ratelimiter to gofri/go-github-ratelimit as recommended by github-go. See https://github.com/google/go-github?tab=readme-ov-file#rate-limiting

WIP: needs testing.


internal/fetchclient/fetchclient.go
```diff
@@ -42,20 +42,21 @@ type Client struct {
 
 // New returns a new client.
 func New(ctx context.Context) *Client {
```
emcfarlane, line 44: Ideally we should alter the `ctx` to set:
```go
// disable go-github's built-in rate limiting
ctx := context.WithValue(context.Background(), github.BypassRateLimitCheck, true)
```

internal/fetchclient/fetchclient.go
```diff
@@ -42,20 +42,21 @@ type Client struct {
...
+	githubToken := os.Getenv("GITHUB_TOKEN")
+	if githubToken != "" {
+		log.Printf("creating authenticated GitHub client with GITHUB_TOKEN")
 	} else {
-		retryableClient := retryablehttp.NewClient()
-		retryableClient.Logger = nil
-		client = retryableClient.StandardClient()
+		log.Printf("creating unauthenticated GitHub client")
 	}
+	rateLimiter, err := github_ratelimit.NewRateLimitWaiterClient(nil)
```
emcfarlane, line 51: From the current code this never errors, so we could ignore too.

mfridman (commented):
This may or may not fix our issues, but I'm not sure how much I like bringing this library as a dependency and giving it full control over the transport.

internal/fetchclient/fetchclient.go
```diff
@@ -41,21 +41,27 @@ type Client struct {
...
+		// Fallback to default client if rate limiter creation fails.
+		ctx = context.WithValue(ctx, github.SleepUntilPrimaryRateLimitResetWhenRateLimited, true)
+		httpClient = http.DefaultClient
+	} else {
+		// Disable the github-go rate limiter for github_ratelimit.
+		ctx = context.WithValue(ctx, github.BypassRateLimitCheck, true)
+	}
+	githubClient := github.NewClient(httpClient).WithAuthToken(githubToken)
+	return ctx, &Client{
+		httpClient: httpClient,
```
emcfarlane, line 63: Need to validate this `httpClient` doesn't require oauth2 tokens.

emcfarlane: `httpClient` looks to only be called for fetching against other registries. Should _not_ have the token `GITHUB_TOKEN` set.

internal/fetchclient/fetchclient.go
```diff
@@ -41,21 +41,21 @@ type Client struct {
...
 // New returns a new client.
-func New(ctx context.Context) *Client {
-	var client *http.Client
-	if ghToken := os.Getenv("GITHUB_TOKEN"); ghToken != "" {
-		ts := oauth2.StaticTokenSource(
-			&oauth2.Token{AccessToken: ghToken},
-		)
-		client = oauth2.NewClient(ctx, ts)
+func New(ctx context.Context) (context.Context, *Client) {
+	httpClient := http.DefaultClient
```
mfridman, line 45: Something I found useful in the past is checking the status of the rate limits after instantiating the client. I wonder if we're just exhausting the amount of work that a bot user can do across our repositories.

https://docs.github.com/en/rest/using-the-rest-api/rate-limits-for-the-rest-api?apiVersion=2022-11-28#checking-the-status-of-your-rate-limit

The only downside is there's no way (afaik) to check the secondary rate limits, and if we're hitting those then 🤷 .

emcfarlane:
Closing, will reopen when tested.

PR #1843, opened 2025-05-20
Switch to buf-action to setup buf CLI

pkwarren:


.github/workflows/arm64.yml
```diff
@@ -41,9 +41,9 @@ jobs:
       shell: bash
       run: |
         echo BUF_VERSION=$(go list -m -f '{{.Version}}' github.com/bufbuild/buf) >> $GITHUB_ENV
-    - name: Install buf cli
-      run: |
-        go install github.com/bufbuild/buf/cmd/buf@${BUF_VERSION}
+    - uses: bufbuild/buf-action@v1
```
mfridman, line 44: Does this respect `BUF_VERSION`?

mfridman: Looks like it, https://github.com/bufbuild/buf-action?tab=readme-ov-file#versioning

pkwarren: Fixed in 321e12d (although it might be nice if the action could strip a leading 'v' if found).

mfridman (approved):

PR #1874, opened 2025-06-11
Add support for apple/servicetalk plugin

### LINKED ISSUE #1865: Plugin request for Buf Schema Registry - apple/servicetalk
## Mandatory

https://github.com/apple/servicetalk

Specifically the https://github.com/apple/servicetalk/tree/main/servicetalk-grpc-protoc

## Optional

Yes, currently on `0.42.56` with nice changelog

https://github.com/apple/servicetalk/releases/tag/0.42.56

Not sure.

Not sure.

pkwarren:
Fixes #1865.

### DIFF
--- a/tests/plugins_test.go
+++ b/tests/plugins_test.go
@@ -2,6 +2,7 @@ package tests
 
 import (
 	"bufio"
+	"bytes"
 	"cmp"
 	"context"
 	"encoding/xml"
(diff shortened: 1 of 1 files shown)


mfridman:
Awesome work!

plugins/apple/servicetalk/source.yaml
```diff
@@ -0,0 +1,4 @@
+source:
```
mfridman, line 1: Oh I sometimes forget about this file. I wonder if we should have a quick lint to fail if a directory doesn't contain this file.

mfridman (approved):

PR #1883, opened 2025-06-13
Use GitHub client instead of http client

pkwarren:
Use the higher level GitHub client to make requests to ensure we correctly handle rate limit errors.

### DIFF
--- a/cmd/download-plugins/main.go
+++ b/cmd/download-plugins/main.go
@@ -188,20 +188,12 @@ func downloadReleaseToDir(ctx context.Context, client *github.Client, plugin rel
 	if err != nil {
 		return fmt.Errorf("failed to make HTTP request: %w", err)
 	}
-	resp, err := client.BareDo(ctx, req)
-	if err != nil {
-		return fmt.Errorf("failed to perform HTTP request: %w", err)
-	}
-	defer func() {
-		if err := resp.Body.Close(); err != nil {
-			log.Printf("failed to close response: %v", err)
-		}
-	}()
 	digest := sha256.New()
 	w := io.MultiWriter(f, digest)
-	if _, err := io.Copy(w, resp.Body); err != nil {
-		return fmt.Errorf("failed to copy file: %w", err)
+	if _, err := client.Do(ctx, req, w); err != nil {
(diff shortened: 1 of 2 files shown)


stefanvanburen (approved):

cmd/download-plugins/main.go
```diff
@@ -184,11 +184,11 @@ func downloadReleaseToDir(ctx context.Context, client *http.Client, plugin relea
 		}
 	}()
 	log.Printf("downloading: %v", plugin.URL)
-	req, err := http.NewRequestWithContext(ctx, http.MethodGet, plugin.URL, nil)
+	req, err := client.NewRequest(http.MethodGet, plugin.URL, nil)
 	if err != nil {
 		return fmt.Errorf("failed to make HTTP request: %w", err)
 	}
-	resp, err := client.Do(req)
+	resp, err := client.BareDo(ctx, req)
```
mfridman, line 191: Do we have to handle this now?

> If rate limit is exceeded and reset time is in the future, BareDo returns *RateLimitError immediately without making a network API call.

pkwarren: I think this is only if we would've ignored a previous rate limit in another request.

With `SleepUntilPrimaryRateLimitResetWhenRateLimited` enabled, it should wait until the request can be fulfilled so we won't have to worry about this.

mfridman: Okay, I guess I don't get why do _both_ changes?

Afaics we've done what is expected in the previous PR, i.e., relating to https://github.com/google/go-github/pull/3117

pkwarren: If you look at the tests from that PR, it is using the GitHub client, not the http.Client. I believe if we make calls on the http.Client, we miss some of these behaviors (handling errors, retrying requests, etc.).

mfridman: Sure, my only question is why switching from `.Do` to `.BareDo`

pkwarren: We could use `.Do` just would have to move around the writer, remove the body close, etc. I can do that if you'd prefer.

mfridman (approved):

PR #1933, opened 2025-07-31
Add support for connectrpc/python

stefanvanburen:
Starting with v0.4.2, the initial/latest version.

### DIFF
--- a/.github/docker/Dockerfile.python
+++ b/.github/docker/Dockerfile.python
@@ -1,3 +1,3 @@
-FROM python:3.11.13-alpine3.22
+FROM python:3.11.13-bookworm
 
 CMD echo this is a dummy file used to automate dependency upgrades for plugins
--- a/plugins/connectrpc/python/v0.4.2/Dockerfile
+++ b/plugins/connectrpc/python/v0.4.2/Dockerfile
@@ -1,0 +1,15 @@
+# syntax=docker/dockerfile:1.15
+FROM python:3.11.13-bookworm AS build
+WORKDIR /app
+RUN python -mvenv /app
+ADD /requirements.txt requirements.txt
+RUN . ./bin/activate \
+ && pip install --no-cache-dir "connect-python[compiler]" \
+ && pip uninstall --yes pip setuptools \
+ && rm -f requirements.txt bin/activate.fish bin/activate.csh bin/Activate.ps1 \
+ && rm -f bin/protoc-gen-connect_python
+
+FROM python:3.11.13-bookworm
+COPY --from=build --link /app /app
+USER nobody
+ENTRYPOINT [ "/app/bin/protoc-gen-connect_python" ]
--- a/plugins/connectrpc/python/v0.4.2/Dockerfile
+++ b/plugins/connectrpc/python/v0.4.2/Dockerfile
@@ -6,8 +6,7 @@ ADD /requirements.txt requirements.txt
 RUN . ./bin/activate \
  && pip install --no-cache-dir -r requirements.txt \
  && pip uninstall --yes pip setuptools \
- && rm -f requirements.txt bin/activate.fish bin/activate.csh bin/Activate.ps1 \
- && rm -f bin/protoc-gen-connect_python
+ && rm -f requirements.txt bin/activate.fish bin/activate.csh bin/Activate.ps1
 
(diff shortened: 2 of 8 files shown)


plugins/connectrpc/python/v0.4.2/requirements.txt
```diff
@@ -0,0 +1,14 @@
...
+connect-python==0.4.2
+frozenlist==1.7.0
+idna==3.10
+multidict==6.6.3
+propcache==0.3.2
+protobuf==6.31.1
+protogen==0.3.1
+typing_extensions==4.14.1
+urllib3==2.5.0
+yarl==1.20.1
```
stefanvanburen, line 14: for now, this includes all the deps of the entire connect-python, since the "compiler" (read: generator) is just an extra of the package.

I think we could probably trim this down, because the [plugin][1] imports the [generator][2], but otherwise seems self contained.

Generated this following the steps here: https://github.com/bufbuild/plugins/pull/688#discussion_r1279637498

[1]: https://github.com/connectrpc/connect-python/blob/v0.4.2/src/connectrpc/protoc_gen_connect_python/main.py
[2]: https://github.com/connectrpc/connect-python/blob/v0.4.2/src/connectrpc/generator.py

.github/docker/Dockerfile.python
```diff
@@ -1,3 +1,3 @@
-FROM python:3.11.13-alpine3.22
+FROM python:3.11.13-bookworm
```
stefanvanburen, line 1: chatted with @pkwarren about switching this from alpine, because debian is what distroless uses as a base.

plugins/connectrpc/python/v0.4.2/buf.plugin.yaml
```diff
@@ -0,0 +1,17 @@
+version: v1
+name: buf.build/connectrpc/python
+plugin_version: v0.4.2
+source_url: https://github.com/connectrpc/connect-python
+description: Generates client and server stubs for Connect Python. Compatible with the Connect RPC protocol.
```
stefanvanburen, line 5: no integration_url yet, because we [haven't moved the docs over to connectrpc.com](https://github.com/connectrpc/connect-python/issues/16).

plugins/connectrpc/python/v0.4.2/Dockerfile
```diff
@@ -0,0 +1,15 @@
+# syntax=docker/dockerfile:1.15
+FROM python:3.11.13-bookworm AS build
+WORKDIR /app
+RUN python -mvenv /app
+ADD /requirements.txt requirements.txt
+RUN . ./bin/activate \
+ && pip install --no-cache-dir "connect-python[compiler]" \
+ && pip uninstall --yes pip setuptools \
+ && rm -f requirements.txt bin/activate.fish bin/activate.csh bin/Activate.ps1 \
+ && rm -f bin/protoc-gen-connect_python
```
stefanvanburen, line 10:
```suggestion
 && rm -f requirements.txt bin/activate.fish bin/activate.csh bin/Activate.ps1
```

.github/docker/Dockerfile.python
pkwarren: Thanks for making this change. If/when other python plugins are updated, we might need to tweak their Dockerfiles but this will be good to have build and runtime be on the same Linux base.

plugins/connectrpc/python/v0.4.2/buf.plugin.yaml
```diff
@@ -0,0 +1,17 @@
...
+spdx_license_id: Apache-2.0
+license_url: https://github.com/connectrpc/connect-python/blob/v0.4.2/LICENSE
+output_languages:
+  - python
+registry:
```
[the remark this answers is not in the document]
stefanvanburen: https://github.com/bufbuild/plugins/pull/1933/commits/ba9ab4ea09087de5cb8f54cd5e75152e1bb1fe3f added one on pb/python, which depends on pb/pyi

stefanvanburen:
I think this is all set - tried it out locally; ran into [one issue with dependencies](https://github.com/connectrpc/connect-python/pull/21) that I reported upstream (but is trivially fixed by users by adding the dep manually for now).

PR #1987, opened 2025-08-26
Add test for resolving Python SDK dependencies

stefanvanburen:
Makes sure that we can install the deps by adding them all to a throwaway `uv` project. (Would be nice if we could just hit the PyPI API, but there's not really a way to do that since we have dependency specifiers, and it's also nice to make sure that our specified versions don't conflict with each other when they're eventually resolved into someone's real project.)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -61,7 +61,7 @@ jobs:
     - uses: bufbuild/buf-action@v1
       with:
         setup_only: true
-    - uses: astral-sh/setup-uv@v6
+    - uses: astral-sh/setup-uv@4959332f0f014c5280e7eac8b70c90cb574c9f9b # v6.6.0
     - name: Login to Docker Hub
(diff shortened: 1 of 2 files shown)


stefanvanburen:
Checked that this worked in CI by using `allPlugins`; switched back to only running on the relevant plugins in the PR [here](https://github.com/bufbuild/plugins/pull/1987/commits/0f181ba9b791e4fa3739a311978036fdeffb7f56).

.github/workflows/ci.yml
```diff
@@ -61,6 +61,7 @@ jobs:
     - uses: bufbuild/buf-action@v1
       with:
         setup_only: true
+    - uses: astral-sh/setup-uv@v6
```
pkwarren, line 64: Can you pin this action? I've used https://github.com/suzuki-shunsuke/pinact to do so previously.

stefanvanburen: did it with:

```console
$ pinact run -e "actions/.*" -e "bufbuild/.*"
```

since we trust `actions/` & `bufbuild/` actions.

PR #2000, opened 2025-09-02
Update grpc-web to v2.0.0

### LINKED ISSUE #1999: Unexpected Fetch latest versions failure
The automatic Fetch latest versions workflow failed, please investigate:
https://github.com/bufbuild/plugins/actions/runs/17407258023/attempts/1

pkwarren:
Update the grpc-web plugin to 2.0.0. Unfortunately, we can no longer use prebuilt binaries as the release binaries are created on Ubuntu 24.04, which doesn't work on Debian bookworm based images. When distroless supports Debian Trixie (13), we may be able to switch back to use prebuilt binaries.

Fixes #1999.


plugins/grpc/web/v2.0.0/buf.plugin.yaml
```diff
@@ -0,0 +1,24 @@
+version: v1
+name: buf.build/grpc/web
+plugin_version: v2.0.0
+source_url: https://github.com/grpc/grpc-web
+integration_guide_url: https://grpc.io/docs/platforms/web/quickstart
+description: gRPC for Web Clients.
+deps:
+  - plugin: buf.build/protocolbuffers/js:v3.21.4
```
pkwarren, line 8: Didn't bump this to 4.0.0 yet - the upstream examples show 3.21.4: https://github.com/grpc/grpc-web/blob/875e2baadff1371a514d805c43a815e95355d499/net/grpc/gateway/examples/helloworld/package.json#L10

stefanvanburen (approved):

PR #2029, opened 2025-09-17
Add grpc/web v2.0.2

pkwarren:
The grpc/web plugin now has a corresponding GitHub release so we can update to it.


plugins/grpc/web/v2.0.2/buf.plugin.yaml
```diff
@@ -0,0 +1,24 @@
+version: v1
+name: buf.build/grpc/web
+plugin_version: v2.0.2
+source_url: https://github.com/grpc/grpc-web
+integration_guide_url: https://grpc.io/docs/platforms/web/quickstart
+description: gRPC for Web Clients.
+deps:
+  - plugin: buf.build/protocolbuffers/js:v3.21.4
```
mfridman, line 8: Does this need a bump to v4.0.0?

pkwarren: I was waiting till this was upgraded before bumping: https://github.com/grpc/grpc-web/blob/9e6bf0f521ebeecf7cdde62ee10da289ef8d9b0a/packages/grpc-web/package.json#L33

PR #2108, opened 2025-11-18
Pin grpcio minimum in latest grpc/python version

stefanvanburen:
gRPC's python runtime (`grpcio`) [inserts a runtime check](https://github.com/grpc/grpc/blob/7d256a4a7796e3447854938216a23727913030f1/src/compiler/python_generator.cc#L738-L781) to ensure that the runtime is the same version or newer than the generated plugin code. (This is unfortunately undocumented other than in code, hence why we haven't had this minimum included before.)


stefanvanburen:
(parking this until we know if we want to backport this.)

pkwarren (approved):

PR #2148, opened 2025-12-11
Update ts-proto to last stable 1.x version

NAlexandrov:
Added the lastest stable 1.x version of ts-proto


NAlexandrov:
> Is there a reason you are unable to use a later version of the plugin?

The 2.x release of ts-proto has breaking changes: `ts-proto migrated the low-level Protobuf serializing` and we can't upgrade to the new version due to huge changes in types in TypeScript across the project. Our project is in maintenance mode and will remain in operation for at least three more years. In this mode, we cannot make the significant changes required for the transition to the new version of ts-proto.

Take a look at the number of downloads of ts-proto 1.181.2 over the last 7 days on npm (242 925 vs 101 119 of the latest release). A lot of projects remained on 1.x version.

PR #2164, opened 2025-12-19
feat(plugins): add kratos-http community plugin for v2.9.2

tonybase:


plugins/go-kratos/http/v2.9.2/Dockerfile
```diff
@@ -0,0 +1,15 @@
+# syntax=docker/dockerfile:1.17
+FROM --platform=$BUILDPLATFORM golang:1.25.0-bookworm AS build
+
+ARG TARGETOS TARGETARCH
+ENV CGO_ENABLED=0 GOOS=$TARGETOS GOARCH=$TARGETARCH
+
+RUN --mount=type=cache,target=/go/pkg/mod \
+    go install -ldflags="-s -w" -trimpath go install github.com/go-kratos/kratos/cmd/protoc-gen-go-http/v2@b9fab9a \
```
pkwarren: This should use a tagged version - it will require the upstream project to create a tag named `cmd/protoc-gen-go-http/v2.X.Y` (where `X.Y` are the minor/patch version).

See the grpc-go plugin for an example: https://github.com/grpc/grpc-go/releases/tag/cmd%2Fprotoc-gen-go-grpc%2Fv1.6.0

plugins/go-kratos/http/source.yaml
```diff
@@ -0,0 +1,4 @@
+source:
+  github:
+    owner: go-kratos
+    repository: kratos
```
pkwarren: Once a tagged release is available, this should update to use a goproxy source similar to https://github.com/bufbuild/plugins/blob/main/plugins/grpc/go/source.yaml.

pkwarren (commented):
Thank you for the contribution. It looks like there are some changes needed in the go-kratos project in order to adopt this.

After moving the plugin to the correct location, you can test building the plugin with `make test PLUGINS=community/kratos-http`. The first time this runs it will create `plugin.sum` files used to verify deterministic output from the plugin. These will need to be checked in to get the build passing. See https://github.com/bufbuild/plugins/blob/main/CONTRIBUTING.md#creating-a-new-plugin for more information.

pkwarren:
Closing for now due to inactivity.

PR #2242, opened 2026-02-06
Automate updating plugin versions

stefanvanburen:
First step towards removing some of the work we do when updating plugin versions. This simply looks for the latest plugin version that already exists in the plugins directory, and bumps the `deps:` key to the latest version for plugins.

I _don't_ think this handles the case where the update to a dep is in the same PR as the update to the consuming plugin (I think the syncer can handle that case?), so we could maybe extend this further to detect that situation.

I think we can also go further and start to look for updates to certain deps (e.g., a gRPC runtime package) on registries.

### DIFF
--- a/internal/cmd/fetcher/main.go
+++ b/internal/cmd/fetcher/main.go
@@ -16,10 +16,13 @@ import (
 	"time"
 
 	"buf.build/go/interrupt"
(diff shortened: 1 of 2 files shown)


pkwarren (approved):
Nice!

PR #2280, opened 2026-02-24
Bundle maven packaging deps in java plugins

DMarby:
This updates our fetcher, and adds a standalone tool to render/re-generate Dockerfiles as well, to make them contain the dependencies we need to do maven packaging for a plugin at a known path.

Updates the latest version of each plugin, follow-up PRs will update older versions once we're happy with this approach. Follow-up PR (that likely needs breaking down further) to update all older versions: https://github.com/bufbuild/plugins/pull/2281

### DIFF
--- a/internal/maven/pom_test.go
+++ b/internal/maven/pom_test.go
@@ -222,26 +222,3 @@ registry:
 		})
 	}
 }
-
-func TestXMLEscape(t *testing.T) {
-	t.Parallel()
-	tests := []struct {
-		input    string
-		expected string
-	}{
-		{"normal-text", "normal-text"},
-		{"<tag>", "&lt;tag&gt;"},
-		{"a&b", "a&amp;b"},
-		{`"quoted"`, "&#34;quoted&#34;"},
-		{"'single'", "&#39;single&#39;"},
-		{"<>&\"'", "&lt;&gt;&amp;&#34;&#39;"},
-	}
-	for _, tt := range tests {
-		t.Run(tt.input, func(t *testing.T) {
-			t.Parallel()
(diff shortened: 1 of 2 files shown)


stefanvanburen (commented):
seems like a reasonable approach to me, assuming we'd test this locally e2e first. just a couple small thoughts

internal/maven/deps.go
```diff
@@ -0,0 +1,133 @@
...
+	var result []bufremotepluginconfig.MavenDependencyConfig
+	for _, dep := range deps {
+		key := dep.GroupID + ":" + dep.ArtifactID
+		if dep.Classifier != "" {
+			key += ":" + dep.Classifier
+		}
+		if existingVersion, ok := seen[key]; ok {
+			if existingVersion != dep.Version {
+				return nil, fmt.Errorf(
+					"duplicate Maven dependency %s with conflicting versions: %s vs %s",
```
pkwarren: This isn't actually an error - we ideally want to duplicate this behavior of how Maven resolves transitive dependencies: https://maven.apache.org/guides/introduction/introduction-to-dependency-mechanism.html#transitive-dependencies

The algorithm should be:
1. Add all dependencies in the child plugin (from all runtimes). This is ok to fail if there's a conflict.
2. Find all transitive dependencies pulled in by the plugin's recursive deps. For each dependency not included in the child plugin, pick the nearest definition.

For a different approach (that may include more than we need but be easier to reason about), we could have the pom contain multiple profiles - one per plugin. Then run `dependency:go-offline` against each profile in the Dockerfile, which would ensure that all dependencies of the child plugin and all its dependencies are included in the repository. This would let Maven do the heavy lifting.

PR #2300, opened 2026-03-03
Allow configuring an update frequency for plugins

pkwarren:
Update the source.yaml file to support and `update_frequency` setting. This is used for frequently updated plugins that we had disabled in the past (as there were a large number of plugin updates without always changing the generated code). For now, set all of the previously disabled plugins to prevent creating new versions for 30 days after the last plugin update.

The last plugin update is calculated by doing a full Git checkout and using `git log` to determine the create time from the latest version of the plugin.

### DIFF
--- a/plugins/community/chrusty-jsonschema/source.yaml
+++ b/plugins/community/chrusty-jsonschema/source.yaml
@@ -1,5 +1,5 @@
 source:
-  update_frequency: 30d
+  disabled: true
   github:
     owner: chrusty
     repository: protoc-gen-jsonschema


internal/cmd/fetcher/main.go
```diff
@@ -472,50 +569,45 @@ func run(ctx context.Context, container appext.Container, fetcher Fetcher, f *fl
...
+	if createTime.IsZero() {
+		return false, nil
+	}
+	freq := time.Duration(*config.Source.UpdateFrequency)
+	nextUpdate := createTime.Add(freq)
+	if time.Now().Before(nextUpdate) {
+		logger.InfoContext(ctx, "skipping source (update frequency not reached)",
+			slog.String("filename", config.Filename),
+			slog.String("last_updated", createTime.UTC().Format(time.DateOnly)),
+			slog.String("next_update", nextUpdate.UTC().Format(time.DateOnly)),
```
pkwarren: Here's an example log (manually removed latest mercari plugin and set `1000d` as the interval in the config):
```
time=2026-03-03T15:59:24.323-06:00 level=INFO msg="skipping source (update frequency not reached)" filename=plugins/community/mercari-grpc-federation/source.yaml last_updated=2025-03-24 next_update=2027-12-19
```

This matches the time when it had last been updated: https://github.com/bufbuild/plugins/pull/1780

PR #2333, opened 2026-03-17
Add anthropics/buffa plugin v0.5.2

iainmcgin:
Adds the `protoc-gen-buffa` plugin for Rust protobuf code generation, under the new top-level `anthropics` org as requested in earlier review.

## About buffa

[buffa](https://github.com/anthropics/buffa) is a zero-copy Rust Protobuf implementation with:
- **Editions support** (2023/2024) and full proto2/proto3 coverage
- **Zero-copy view types** - borrowed `&str`/`&[u8]` field access without owned allocation
- **no_std + alloc** compatible
- **Proto3 JSON mapping** with serde integration
- **Conformance**: passes the protobuf conformance suite (5539 binary+JSON tests, 0 expected failures)

Crates are on crates.io: [`buffa`](https://crates.io/crates/buffa), [`buffa-types`](https://crates.io/crates/buffa-types), [`protoc-gen-buffa`](https://crates.io/crates/protoc-gen-buffa).

## Output convention

With `file_per_package=true` (set in `opts`), the plugin emits **one `<dotted.package>.rs` file per proto package**: `foo/v1/bar.proto` and `foo/v1/baz.proto` both land in `foo.v1.rs`. No subdirectories, no `mod.rs`. This matches the prost/tonic filename convention the BSR Rust SDK `lib.rs` synthesis already understands, so no special-casing should be required on the BSR side.

`plugin.sum` files for `eliza` and `petapis` are checked in (`make test PLUGINS=anthropics/buffa`).

Related: #2334 (anthropics/connect-rust) declares this plugin as a dep, so this one needs to land first.


plugins/anthropics/buffa/v0.3.0/Dockerfile
```diff
@@ -0,0 +1,15 @@
+# syntax=docker/dockerfile:1.19
+FROM rust:1.91.1-alpine3.22 AS builder
```
pkwarren, line 2: We should update this branch to latest main then we can update this image to the latest base image:

```suggestion
FROM rust:1.95.0-alpine3.23@sha256:606fd313a0f49743ee2a7bd49a0914bab7deedb12791f3a846a34a4711db7ed2 AS builder
```

plugins/anthropics/buffa/v0.3.0/Dockerfile
```diff
@@ -0,0 +1,15 @@
...
+RUN apk add --no-cache musl-dev
+WORKDIR /app
+ENV CARGO_REGISTRIES_CRATES_IO_PROTOCOL=sparse
+RUN --mount=type=cache,target=/usr/local/cargo/registry,sharing=locked --mount=type=cache,target=/root/target \
+    cargo install protoc-gen-buffa --version 0.3.0 --locked --root /app
+
+FROM gcr.io/distroless/static-debian12:latest@sha256:87bce11be0af225e4ca761c40babb06d6d559f5767fbf7dc3c47f0f1a466b92c AS base
```
pkwarren, line 9:
```suggestion
FROM gcr.io/distroless/static-debian13:latest@sha256:47b2d72ff90843eb8a768b5c2f89b40741843b639d065b9b937b07cd59b479c6 AS base
```

plugins/anthropics/buffa/v0.3.0/buf.plugin.yaml
```diff
@@ -0,0 +1,41 @@
...
+plugin_version: v0.3.0
+source_url: https://github.com/anthropics/buffa
+description: Generates Rust message types with buffa, a zero-copy Protobuf implementation with editions support and no_std compatibility.
+output_languages:
+  - rust
+spdx_license_id: Apache-2.0
+license_url: https://github.com/anthropics/buffa/blob/v0.3.0/LICENSE
+registry:
+  cargo:
+    rust_version: "1.85"
```
pkwarren, line 12: Is this the minimum supported version?

pkwarren (commented):
To ensure reproducible builds, we check in a checksum file of generated code using the plugin. This verifies e2e code generation is functional and also ensures that the codegen doesn't change across builds.

https://github.com/bufbuild/plugins/blob/main/CONTRIBUTING.md#creating-a-new-plugin

To create the files, run `make test PLUGINS=anthropics/buffa:latest` and check in the `plugin.sum` files created under `tests/testdata/buf.build/anthropics/`.

plugins/anthropics/buffa/v0.3.0/buf.plugin.yaml
pkwarren: Answered my own question: https://github.com/anthropics/buffa/blob/fcb490109f9b76746815f0534211856bf2d7baf0/Cargo.toml#L27

stefanvanburen:
> 1. **opts**: I set `json=true` and `views=true`. If BSR has conventions for which opts to expose vs bake in, please advise.

Seems reasonable to me to enable both — since generated SDKs aren't configurable, I think including all of the features makes sense to me (users who want more control can fall back to local generation).

> 2\. **File naming**: The per-file output naming (`pkg.subpkg.file.rs`) is flat but per-proto-file, not per-package like prost. If BSR's lib.rs synthesis expects per-package naming instead, we can add a plugin option to switch modes.

What would be _most_ helpful here is actually if both buffa and connect-rust included the `lib.rs` generation logic themselves (enabled with a plugin option; `generate_librs=true` strawman name; disabled by default) so that the BSR doesn't need any sort of special-casing for generating it on our end — is that possible to do?

iainmcgin:
@stefanvanburen buffa has a plugin for local usage, [protoc-gen-buffa-packaging](https://github.com/anthropics/buffa#using-buf-generate-recommended), which produces a consolidated mod.rs separately from the core plugin. This was the structure that @bufdev wanted me to use because that plugin requires `strategy: all` to get visibility across all files in one pass. I'm not sure if we can generate `lib.rs` without `strategy: all` for similar reasons, but open to suggestions.

iainmcgin:
[claude code] We'll add a `file_per_package=true` opt to `protoc-gen-buffa` that emits one `<dotted.package>.rs` per proto package (concatenating what currently goes into the per-proto-file outputs). That matches the convention the existing cargo-SDK `lib.rs` synthesis already handles for prost, so no special-casing should be needed on the BSR side. It's also correct under `strategy: directory` for any `PACKAGE_DIRECTORY_MATCH`-clean module, so it avoids the `strategy: all` dependency @bufdev flagged earlier.

connect-rust will get the same opt and also emit `<dotted.package>.rs` — since the buffa and connect-rust SDKs are separate crates there's no collision, and the service stubs land at the proto package path (`<connect_sdk>::pet::v1::PetServiceClient`) rather than under an extra submodule the way tonic's `<pkg>.tonic.rs` ends up.

Targeting v0.4.1 for both. I'll update these PRs once that's released.

Separately: as far as I can tell `registry.cargo` only exposes `rust_version` and `deps`, so the filename→module-tree synthesis is an implicit contract that plugins have to reverse-engineer from prost's output. Would it be worth either documenting that convention in CONTRIBUTING.md, or adding something like `registry.cargo.lib_rs: plugin` so a plugin can opt into emitting its own `lib.rs` and have BSR use it verbatim? Happy to open a separate issue if that's a better place for it.

stefanvanburen:
> We'll add a `file_per_package=true` opt to `protoc-gen-buffa` that emits one `<dotted.package>.rs` per proto package (concatenating what currently goes into the per-proto-file outputs).

@iainmcgin sounds good, thanks for your persistence with this, it's appreciated!

> Separately: as far as I can tell `registry.cargo` only exposes `rust_version` and `deps`, so the filename→module-tree synthesis is an implicit contract that plugins have to reverse-engineer from prost's output. Would it be worth either documenting that convention in CONTRIBUTING.md, or adding something like `registry.cargo.lib_rs: plugin` so a plugin can opt into emitting its own `lib.rs` and have BSR use it verbatim? Happy to open a separate issue if that's a better place for it.

You're basically on-the-nose here, unfortunately: with basically every other package ecosystem, it's relatively straightforward to package up plugin output; Rust is a different story, and the BSR has some workarounds here to handle `prost` / `tonic` as a special snowflake, given that they were the only real game in town when we built out Rust generated SDKs. Now that there's more to the ecosystem, we'll hopefully be able to abstract that away a little more.

iainmcgin:
[claude code] Updated to `protoc-gen-buffa` v0.5.2 with `file_per_package=true` set in `opts`. The plugin now emits one `<dotted.package>.rs` per proto package — `eliza` produces `buf.connect.demo.eliza.v1.rs`, `petapis` produces `pet.v1.rs` — matching the prost filename convention. `plugin.sum` files are checked in (`make test PLUGINS=anthropics/buffa` passes locally).

@stefanvanburen following up from [the earlier discussion](https://github.com/bufbuild/plugins/pull/2333#issuecomment-4332235677) — `file_per_package` landed; if the BSR `lib.rs` synthesis recognizes this output shape with no further special-casing, this should be good to go from my side. Let me know if you'd like anything else changed.

pkwarren:
Taking a look at the latest changes today. Will let you know if there are any further changes necessary to work with gen SDKs.

iainmcgin:
[claude code] Superseded by #2334, which now contains both `anthropics/buffa` and `anthropics/connect-rust` and can land standalone. Closing this one to keep review in one place.

PR #2334, opened 2026-03-17
Add anthropics/buffa and anthropics/connect-rust plugins

iainmcgin:
Adds the `protoc-gen-buffa` and `protoc-gen-connect-rust` plugins for Rust protobuf and ConnectRPC code generation, under the new top-level `anthropics` org as requested in earlier review.

This PR contains both plugins, so it can land on its own. (It was originally stacked on #2333, which adds `anthropics/buffa` alone — #2333 is now superseded by this PR.)

## buffa

[buffa](https://github.com/anthropics/buffa) is a zero-copy Rust Protobuf implementation with:
- **Editions support** (2023/2024) and full proto2/proto3 coverage
- **Zero-copy view types** - borrowed `&str`/`&[u8]` field access without owned allocation
- **no_std + alloc** compatible
- **Proto3 JSON mapping** with serde integration
- **Conformance**: passes the protobuf conformance suite (5539 binary+JSON tests, 0 expected failures)

Crates on crates.io: [`buffa`](https://crates.io/crates/buffa), [`buffa-types`](https://crates.io/crates/buffa-types), [`protoc-gen-buffa`](https://crates.io/crates/protoc-gen-buffa).

`protoc-gen-buffa` generates message types. With `file_per_package=true` (set in `opts`), it emits **one `<dotted.package>.rs` file per proto package**: `foo/v1/bar.proto` and `foo/v1/baz.proto` both land in `foo.v1.rs`. No subdirectories, no `mod.rs`. This matches the prost/tonic filename convention the BSR Rust SDK `lib.rs` synthesis already understands.

## connect-rust

[connect-rust](https://github.com/anthropics/connect-rust) is a Tower-based Rust implementation of the ConnectRPC protocol:
- **Three protocols**: Connect, gRPC, gRPC-Web - same service impl handles all three
- **Full conformance**: 3600 server tests, 2580/1454/2838 client tests across Connect/gRPC/gRPC-Web
- **Zero-copy request handling** via buffa view types - `request.name` is a `&str` borrow into the decoded buffer
- **Tower integration**: composes with axum, hyper, tower middleware

Crates on crates.io: [`connectrpc`](https://crates.io/crates/connectrpc), [`connectrpc-codegen`](https://crates.io/crates/connectrpc-codegen), [`connectrpc-build`](https://crates.io/crates/connectrpc-build).

[...]

PR #2370, opened 2026-04-02
Add anthropics/buffa-packaging plugin v0.3.0

iainmcgin:
Adds the `protoc-gen-buffa-packaging` plugin for assembling buffa-style per-file output into a single `mod.rs` module tree.

## What this plugin does

[`protoc-gen-buffa-packaging`](https://crates.io/crates/protoc-gen-buffa-packaging) reads the proto package structure (not message/service bodies) and writes a `mod.rs` that `include!`s each generated file at the right module nesting level. It works with any codegen plugin that emits per-file output named via the buffa naming convention (`foo/v1/bar.proto` -> `foo.v1.bar.rs`), which includes [`protoc-gen-buffa`](https://crates.io/crates/protoc-gen-buffa) itself and plugins layered on top of it (like [`protoc-gen-connect-rust`](https://crates.io/crates/connectrpc-codegen)).

## Why register this in BSR

The buffa plugin (#2333) emits per-file output without a mod.rs. For local `buf generate` users, the packaging plugin assembles the module tree so the generated code can be used as `pub mod proto;` from a single `#[path]`. BSR's lib.rs synthesis handles the SDK-crate case differently, but BSR users who pull the plugin into a buf.gen.yaml workflow still need this packaging step.

## Options

- `filter=services` — only include proto files that declare at least one `service`. Useful when packaging output from a service-stub generator that skips files without services.

The plugin requires `strategy: all` (client-side concern in buf.gen.yaml) so it sees the full file set in one invocation.

## No registry block

This plugin emits only `mod.rs` files containing `#[path]` and `pub mod` declarations - no Rust code with cargo deps. So no `registry:` block is needed (same pattern as `community/neoeinstein-prost-crate`).

## Dependencies

Logically pairs with #2333 (anthropics/buffa) and #2334 (anthropics/connect-rust). The packaging plugin works on output from those plugins but does not declare them as plugin deps - it operates on the descriptor set, not on the generated code.

## Process note

CONTRIBUTING.md asks for an issue first - happy to open one if you would prefer to discuss before reviewing. Filed as draft.

### DIFF
--- a/plugins/anthropics/buffa-packaging/source.yaml
+++ b/plugins/anthropics/buffa-packaging/source.yaml
@@ -1,0 +1,3 @@
+source:
+  crates:
+    crate_name: protoc-gen-buffa-packaging
--- a/plugins/anthropics/buffa-packaging/v0.3.0/.dockerignore
+++ b/plugins/anthropics/buffa-packaging/v0.3.0/.dockerignore
@@ -1,0 +1,2 @@
+*
+!Dockerfile
--- a/plugins/anthropics/buffa-packaging/v0.3.0/Dockerfile
+++ b/plugins/anthropics/buffa-packaging/v0.3.0/Dockerfile
@@ -1,0 +1,15 @@
+# syntax=docker/dockerfile:1.19
+FROM rust:1.91.1-alpine3.22 AS builder
+RUN apk add --no-cache musl-dev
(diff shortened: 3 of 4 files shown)


iainmcgin:
Closing as we don't intend this to be used for remote builds

PR #2389, opened 2026-04-10
Separate download/install phases in a couple Dockerfiles

stefanvanburen:
We don't necessarily need to merge _this_ PR, but just wanted to see if we were missing any places here where we could split up these steps for better layering. Found a couple places. We can also just keep these updates in mind for the next fetcher updates.


stefanvanburen:
Going to close this out and we'll address in future PRs.

PR #2406, opened 2026-04-17
Create a 'make package' makefile target

pkwarren:
Add a `make package` makefile target which packages up a plugin in a .zip file similar to the official release. Refactor the release command to move the .zip file creation to a package shared by the release and the new package command.

### DIFF
--- a/internal/docker/build.go
+++ b/internal/docker/build.go
@@ -2,6 +2,7 @@ package docker
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"io/fs"
 	"os"
@@ -22,7 +23,7 @@ func Build(
 	imageName string,
 	cachePath string,
 	args []string,
-) ([]byte, error) {
+) (_ []byte, retErr error) {
 	identity := plugin.Identity
 	commonArgs := []string{
 		"buildx",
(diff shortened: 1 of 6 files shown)


internal/docker/build.go
```diff
@@ -74,12 +75,19 @@ func Build(
...
+	defer func() {
+		retErr = errors.Join(retErr, root.Close())
+	}()
 	if err := filepath.WalkDir(filepath.Dir(plugin.Path), func(path string, d fs.DirEntry, err error) error {
 		if err != nil {
 			return err
 		}
 		if !d.IsDir() {
-			if err := os.Chtimes(path, time.Time{}, time.Now().UTC()); err != nil {
+			if err := root.Chtimes(path, time.Now(), time.Now()); err != nil {
```
stefanvanburen: just double-checking that flipping these from UTC -> local(?) is fine?

stefanvanburen (approved):

internal/docker/build.go
pkwarren: Fixed in 099ebb8e.

stefanvanburen (approved):

stefanvanburen (approved):

PR #2415, opened 2026-04-21
Optimize release command

pkwarren:
The release command is very inefficient today. It currently scans all plugins looking in GHCR for every changed image since the last time it ran. It is possible however to optimize this to use the GHCR package API to check for plugins updated since the last release along with git changes since the last release to only check a handful of plugins instead of all ~1700 versions.

Continue to support a PLUGINS env var as an escape hatch to check selected plugins.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -5,6 +5,11 @@ on:
     # At minute 30 past every 6th hour.
     - cron: "30 */6 * * *"
   workflow_dispatch:
+    inputs:
+      plugins:
+        description: 'Force-include plugins (comma-separated, e.g. "connect-go,grpc/csharp:v1.68.1", or "all")'
+        required: false
+        default: ''
 
 permissions:
   contents: write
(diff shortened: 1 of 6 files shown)


.github/workflows/release.yml
```diff
@@ -5,6 +5,11 @@ on:
     # At minute 30 past every 6th hour.
     - cron: "30 */6 * * *"
   workflow_dispatch:
+    inputs:
+      plugins:
+        description: 'Force-include plugins (comma-separated, e.g. "connect-go,grpc/csharp:v1.68.1", or "all")'
+        required: false
+        default: ''
```
pkwarren, line 12: Added this as a fallback if needed (although local testing this is doing the right thing).

pkwarren (commented):
When running locally I get:

```
time=2026-04-21T15:39:09.239-05:00 level=INFO msg="plugin scan scope" candidates=51 total=1712
```

So even with looking back 30 minutes prior to the last release, we've filtered down what needs to be checked to 51/1712 plugins.

stefanvanburen (approved):

PR #2425, opened 2026-04-22
Use per-package instead of org endpoints

pkwarren:
With the `${github.token}` or an app token, it appears that the org level package endpoints can't be retrieved. It is not worth switching this to a PAT.

Update the release workflow to call each package's endpoints directly (~81 calls vs 1) which is still a marked improvement from before (~1700 OCI registry calls).

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -12,7 +12,7 @@ on:
         default: ''
 
 permissions:
-  contents: read
+  contents: write
   id-token: write
   issues: write
   packages: read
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -26,14 +26,6 @@ jobs:
...
     runs-on: ubuntu-latest
     steps:
-      - name: Generate token
-        id: generate_token
-        uses: actions/create-github-app-token@1b10c78c7865c340bc4f6099eb2f838309f1e8c3 # v3.1.1
-        with:
-          client-id: ${{ secrets.TOKEN_EXCHANGE_GH_APP_CLIENT_ID }}
-          private-key: ${{ secrets.TOKEN_EXCHANGE_GH_APP_PRIVATE_KEY }}
-          permission-contents: write
-          permission-packages: read
```
pkwarren, line 36: Reverted previous attempts to use an app token.

DMarby (approved):

PR #2441, opened 2026-04-28
Add PyPI source type and migrate pip-installed Python plugins

### LINKED ISSUE #2440: Unexpected Fetch latest versions failure
The automatic Fetch latest versions workflow failed, please investigate:
https://github.com/bufbuild/plugins/actions/runs/25060004981/attempts/1

stefanvanburen:
Fixes #2440.

Add a pypi source type to the fetcher so plugins distributed as Python packages can track releases from PyPI rather than GitHub tags. The fetcher queries `https://pypi.org/pypi/{name}/json`, skips ~~yanked~~ (_edit_: yanked out the yanked handling) and pre-release versions, and respects the existing ignore_versions and max_version filters.

Migrate four plugins whose Dockerfiles install via pip to the new source type:
- `connectrpc/python`  (`protoc-gen-connect-python`)
- `community/nipunn1313-mypy`  (`mypy-protobuf`)
- `community/nipunn1313-mypy-grpc`  (`mypy-protobuf`)
- `community/danielgtaylor-betterproto`  (`betterproto`)

Leave `protocolbuffers/python`, `protocolbuffers/pyi`, and `grpc/python` on github, as all three compile C++ from source using Bazel and version against the upstream monorepo tag rather than any PyPI package.


internal/fetchclient/fetchclient.go
```diff
@@ -27,6 +27,8 @@ const (
 	goProxyURL        = "https://proxy.golang.org"
 	npmRegistryURL    = "https://registry.npmjs.org"
 	mavenURL          = "https://repo1.maven.org/maven2"
+	// docs: https://warehouse.pypa.io/api-reference/json.html
```
pkwarren: This link isn't working for me - should it be updated to https://docs.pypi.org/api/json/?

internal/fetchclient/fetchclient.go
```diff
@@ -417,6 +423,67 @@ func (c *Client) fetchGithub(
 	return versions[len(versions)-1], nil
 }
 
+func (c *Client) fetchPyPI(ctx context.Context, name string, ignoreVersions map[string]struct{}, maxVersion string) (string, error) {
+	request, err := http.NewRequestWithContext(
+		ctx,
+		http.MethodGet,
+		fmt.Sprintf("%s/%s/json", c.pypiBaseURL, strings.TrimPrefix(name, "/")),
```
pkwarren: Do we expect PyPI.Name to ever start with `/`?

internal/fetchclient/fetchclient.go
```diff
@@ -417,6 +423,67 @@ func (c *Client) fetchGithub(
...
+	}
+	defer response.Body.Close()
+	if response.StatusCode != http.StatusOK {
+		return "", fmt.Errorf("received status code %d retrieving %q", response.StatusCode, request.URL.String())
+	}
+
+	var data struct {
+		Releases map[string][]struct {
+			Yanked bool `json:"yanked"`
+		} `json:"releases"`
```
pkwarren: The releases key on a project's json output is marked as deprecated: https://docs.pypi.org/api/json/#get-a-project

Should we switch to a different API to get this information?

internal/fetchclient/fetchclient.go
```diff
@@ -417,6 +423,67 @@ func (c *Client) fetchGithub(
...
+		return "", err
+	}
+	var versions []string
+	for version, files := range data.Releases {
+		if len(files) == 0 {
+			continue
+		}
+		yanked := true
+		for _, file := range files {
+			if !file.Yanked {
```
pkwarren: It sounds like the yanked value can be a string or a bool:

> yanked: An optional key which may be either a boolean to indicate if the file has been yanked, or a non empty, but otherwise arbitrary, string to indicate that a file has been yanked with a specific reason. If the yanked key is present and is a truthy value, then it SHOULD be interpreted as indicating that the file pointed to by the url field has been “Yanked” as per [PEP 592](https://peps.python.org/pep-0592/).

https://peps.python.org/pep-0691/#project-detail

pkwarren (commented):
Overall this looks good - just a couple things.

pkwarren (approved):

PR #2474, opened 2026-05-18
Fix a few issues in GHA workflows

pkwarren:


.github/workflows/fetch_versions.yml
```diff
@@ -88,7 +89,7 @@ jobs:
         if: ${{ failure() }}
         env:
           GITHUB_TOKEN: ${{ steps.generate_issues_token.outputs.token }}
-          GITHUB_SERVER_URL: ${ github.server_url }}
+          GITHUB_SERVER_URL: ${{ github.server_url }}
```
pkwarren, line 92: Surprisingly this never broke anything. I guess if there is a syntax error it just doesn't set the env var, falling back to the existing one?

.github/workflows/ci.yml
```diff
@@ -43,8 +43,10 @@ jobs:
       if: ${{ inputs.plugins == '' }}
       env:
         GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        REPOSITORY: ${{ github.repository }}
+        REF_NAME: ${{ github.ref_name }}
```
emcfarlane, line 47: This for consistency?

emcfarlane (approved):

.github/workflows/ci.yml
DMarby: It avoids a fair bit of injection issues, so is generally recommended vs templating the variables directly in a command (our semgrep config in other repos enforces it)

.github/workflows/fetch_versions.yml
```diff
@@ -11,6 +8,10 @@ on:
 permissions:
   contents: read
   issues: write
+
+# Only allow a single fetch-versions workflow to run at a time so a manual run
+# and the scheduled run don't race on the fetch-versions PR branch.
+concurrency: ${{ github.workflow }}
```
DMarby, line 14: TIL there's a short form version of `concurrency`, only knew about

```
concurrency:
  group: ${{ github.workflow }}
  cancel-in-progress: false
```

DMarby (approved):

.github/workflows/ci.yml
pkwarren: More of a security best practice for args passed to the shell.

PR #2497, opened 2026-05-27
Disable protoc-gen-validate plugins

pkwarren:
The PGV repo is archived and there will no longer be any updates to it. Disable looking for new versions in the fetcher.

### DIFF
--- a/plugins/bufbuild/validate-cpp/source.yaml
+++ b/plugins/bufbuild/validate-cpp/source.yaml
@@ -1,5 +1,5 @@
 source:
-  # Use protovalidate instead
+  # Use https://protovalidate.com instead
   disabled: true
   github:
     owner: bufbuild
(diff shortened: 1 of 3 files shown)


plugins/bufbuild/validate-cpp/source.yaml
```diff
@@ -1,4 +1,6 @@
 source:
+  # Use protovalidate instead
```
mfridman: nit, maybe add a link like https://github.com/bufbuild/protovalidate or https://protovalidate.com/?

pkwarren: Will do but most people won't look at the source.yaml files for a plugin (and the plugins are already marked as deprecated).

oliversun9 (approved):

PR #2500, opened 2026-05-30
Fix buf.plugin.yaml for connect-rust v0.6.1

oliversun9:
It actually depends on 0.6.0 of buffa: https://github.com/anthropics/connect-rust/blob/v0.6.1/Cargo.toml#L38-L40


plugins/anthropics/connect-rust/v0.6.1/buf.plugin.yaml
```diff
@@ -4,7 +4,7 @@ plugin_version: v0.6.1
 source_url: https://github.com/anthropics/connect-rust
 description: Generates ConnectRPC service traits, typed clients, and monomorphic dispatchers for Rust. Compatible with the Connect, gRPC, and gRPC-Web protocols. Passes the full ConnectRPC conformance suite.
 deps:
-  - plugin: buf.build/anthropics/buffa:v0.7.0
+  - plugin: buf.build/anthropics/buffa:v0.6.0
```
pkwarren: The dependency is ^0.6 so 0.6.1 is valid here. Let's keep it as 0.6.1 in this file.

pkwarren (approved):

PR #2619, opened 2026-07-21
Remove patch from go.mod go version

pkwarren:
The actions/setup-go GHA will not use the latest security/bugfix version of Go if the patch version is zero (even with `use-latest-version`). To work around this for now, remove the patch version from go.mod which should allow it to run with later versions.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/bufbuild/plugins
 
-go 1.26.0
+go 1.26
 
 require (
 	aead.dev/minisign v0.3.0


pkwarren:
Before:
```
Attempting to resolve the latest version from the manifest...
matching 1.26.0...
Resolved as '1.26.0'
```

After:
```
Attempting to resolve the latest version from the manifest...
matching 1.26...
Resolved as '1.26.5'
```

emcfarlane (approved):

PR #2622, opened 2026-07-22
Move to gcloud storage command

pkwarren:
The gsutil command is deprecated and it is recommended to move to the 'gcloud storage' command instead.

### DIFF
--- a/.github/workflows/upload.yml
+++ b/.github/workflows/upload.yml
@@ -51,7 +51,7 @@ jobs:
         run: |
           go run ./cmd/download-plugins -since "${SINCE}" downloads
       - name: Upload To Release Bucket
-        run: gsutil -m rsync -r downloads gs://buf-plugins
+        run: gcloud storage rsync --recursive downloads gs://buf-plugins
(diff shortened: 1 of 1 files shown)


.github/workflows/upload.yml
```diff
@@ -51,7 +51,7 @@ jobs:
         run: |
           go run ./cmd/download-plugins -since "${SINCE}" downloads
       - name: Upload To Release Bucket
-        run: gsutil -m rsync -r downloads gs://buf-plugins
+        run: gcloud storage rsync --recursive downloads gs://buf-plugins
```
pkwarren, line 54: https://docs.cloud.google.com/storage/docs/gsutil-transition-to-gcloud

mfridman (approved):

.github/workflows/upload.yml
emcfarlane: `--no-ignore-symlinks` seems like a diff, but presume we dont need it!

emcfarlane (approved):

PR #2656, opened 2026-08-11
Add grpc-py v0.3.0 plugin

pkwarren:


plugins/bufbuild/grpc-py/v0.3.0/buf.plugin.yaml
```diff
@@ -0,0 +1,18 @@
...
+license_url: https://github.com/bufbuild/protoc-gen-grpc-py/blob/v0.3.0/LICENSE
+output_languages:
+- python
+registry:
+  python:
+    package_type: "runtime"
+    requires_python: ">=3.10"
+    deps:
+      # https://pypi.org/project/grpcio/
+      - "grpcio>=1.68"
```
oliversun9, line 18: The example project [requires 1.81](https://github.com/bufbuild/protoc-gen-grpc-py/blob/b05e9d2ca06187b3e341e89d2bf014f9bf0b99bf/example/pyproject.toml#L6) and the [latest on grpcio is 1.83](https://pypi.org/project/grpcio/). How do we decide the min version?

pkwarren: We deliberately keep these back to older versions in case users are stuck on an older version for any reason.

oliversun9 (approved):

PR #2675, opened 2026-08-21
Add a more comprehensive python deps test

pkwarren:
Instead of just verifying the versions are valid for an individual plugin, build a directory structure containing a plugin and its dependencies and attempt to resolve everything together. This catches issues like #2674 and also validates the requires-python values. The requires-python validation caught two issues in mypy plugins at the latest version.


plugins/community/nipunn1313-mypy-grpc/v5.1.0/buf.plugin.yaml
```diff
@@ -11,7 +11,8 @@ registry:
   python:
     package_type: "stub-only"
     # https://github.com/nipunn1313/mypy-protobuf/blob/v5.1.0/pyproject.toml#L21
-    requires_python: ">=3.8"
+    # Bumped to match protobuf requirement.
+    requires_python: ">=3.9"
```
pkwarren, line 15: The protobuf version below requires 3.9 so needed to bump these.

emcfarlane (approved):

pkwarren (commented):
The test would've caught the previous issue:

```
--- FAIL: TestPyPIDependencies (1.91s)
    --- FAIL: TestPyPIDependencies/connectrpc/py@v0.11.1 (0.03s)
        plugins_test.go:506:
                Error Trace:    /Users/pkw/dev/plugins/tests/plugins_test.go:506
                Error:          Received unexpected error:
                                exit status 1
                Test:           TestPyPIDependencies/connectrpc/py@v0.11.1
                Messages:       uv lock failed - output: Using CPython 3.13.2
                                  × No solution found when resolving dependencies:
                                  ╰─▶ Because connectrpc>=0.11.1 depends on protobuf-py==0.1.1 and
                                      bufbuild-py==0.2.0 depends on protobuf-py>=0.2.0, we can conclude that
                                      bufbuild-py==0.2.0 and connectrpc>=0.11.1 are incompatible.
                                      And because only bufbuild-py==0.2.0 is available, we can conclude that
                                      all versions of bufbuild-py and connectrpc>=0.11.1 are incompatible.
                                      And because your project depends on bufbuild-py and connectrpc>=0.11.1,
                                      we can conclude that your project's requirements are unsatisfiable.
```

oliversun9 (approved):

PR #2691, opened 2026-08-27
Fix handling registry image 404

oliversun9:
The caller of the modified function:

```
	registryImage, imageID, err := fetchRegistryImageAndImageID(p)
	if err != nil {
		return release.PluginRelease{}, false, err
	}
	if registryImage == "" || imageID == "" {
		c.logger.InfoContext(ctx, "unable to detect registry image and image ID",
			slog.String("owner", identity.Owner()),
			slog.String("plugin", identity.Plugin()),
			slog.String("version", p.PluginVersion),
		)
		return release.PluginRelease{}, true, nil
	}

```

So we need `fetchRegistryImageAndImageID` to be able to return `"", "", nil`


pkwarren:
This is a good fix however the underlying issue is that this CI workflow has been hung for ~22 hours: https://github.com/bufbuild/plugins/actions/runs/32984329655

Unfortunately it is unable to be canceled. We can merge this but I'm going to disable the release workflow until we can get a build going again.

go.sum
```diff
@@ -72,6 +72,8 @@ golang.org/x/sync v0.22.0 h1:SZjpbeLmrCk4xhRSZFNZW5gFUeCeFgjekvI/+gfScek=
 golang.org/x/sync v0.22.0/go.mod h1:9xrNwdLfx4jkKbNva9FpL6vEN7evnE43NNNJQ2LF3+0=
 golang.org/x/sys v0.47.0 h1:o7XGOvZQCADBQQ4Y7VNq2dRWQR7JmOUW8Kxx4ZsNgWs=
 golang.org/x/sys v0.47.0/go.mod h1:4GL1E5IUh+htKOUEOaiffhrAeqysfVGipDYzABqnCmw=
+golang.org/x/tools v0.49.0 h1:3NI7VXzL9+1WZD52Dx2ttoPwD5DWrFGpl9mFZDlmisI=
```
pkwarren: There were no changes to go.mod - does `go mod tidy` clean these up?

oliversun9: Yea, fixed now. They were produced by `go mod tidy`. I made some changes after that but didn't clean these up.

pkwarren (approved):

